PR #57, opened 2019-02-04
Add Tobalaba token support

clementperon:


danzipie:
Nice!
Question: why did you set 
`CHAIN_ID = 0`
in `examples/signTx.py`? I would expect the same value of NetworkId

clementperon:
I didn't want to change the actual behavior. 

If you want to have the correct signature for Tobalaba you can set the Chain_ID to 401697

PR #66, opened 2019-03-26
RSK wallet needs access to generic 44' path.

ortegaalfredo:
The RSK wallet needs access to other paths, I believe this is the way to enable the generic 44' path in the application upload script.


TamtamHero:
Can you be more specific in the path required ? Please link your source.

ortegaalfredo:
> Can you be more specific in the path required ? Please link your source.

After some deliberations, we can use a smaller path, so I will make a new PR. What do you mean by "link your source" ?

PR #67, opened 2019-04-03
RSK wallet needs access to BTC 44'00' path.

ortegaalfredo:
In addition to his own path, the RSK wallet needs access to the BTC 44'00' path, I believe this is the way to enable access to the 44'00' BTC path in the application upload script.


TamtamHero:
That's not the right syntax, it should be ` --path "44'/0'"` instead.
Can you link us to a document that explicit why RSK needs to access this path ?

PR #68, opened 2019-04-12
RSK wallet needs access to BTC 44'/00' path.

ortegaalfredo:
In addition to his own path, the RSK wallet needs access to the BTC 44'/00' path.

You can see that in the official documentation here: https://github.com/rsksmart/rskj/wiki/Peg-using-ledger
The first step is "Set a derivation path in the key derivationPath in config.json file. You should use BTC mainnet/testnet derivation path 44'/0'/0'/0/0.". So the RSK wallet need access to this path aswell, for the Peg mechanism of RSK.


TamtamHero:
It sounds pretty arbitrary, why wouldn't it work if you picked the path 44'/137'/0'/0/0 back and forth?

alfred-gw:
Hi, sorry but it wouldn't work with just that path. The explanation is a little long, is related to the way RSK generates wallets and can be found in this document:

https://github.com/rsksmart/rskj/wiki/Send-funds-to-RSK

Hope this answers your question.

PR #80, opened 2019-08-26
Fix: Add ETH derivation path to Wanchain 

tyrion70:
Add ETH derivation path to Wanchain to allow users to recover assets sent to ETH address. We've had multiple requests from people that used the ETH derivation path before the ledger live change to fix the derivation path. To recover those assets we would like to allow the ETH derivation path as well on MyWanWallet together with the ledger.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,8 @@ APP_LOAD_PARAMS += --path "44'/200625'"
 DEFINES += CHAINID_UPCASE=\"AKA\" CHAINID_COINNAME=\"AKA\" CHAIN_KIND=CHAIN_KIND_AKROMA CHAIN_ID=200625
 APPNAME = "Akroma"
 else ifeq ($(CHAIN),wanchain)
-APP_LOAD_PARAMS += --path "44'/5718350'"
+# Also allows Wanchain to access the ETH derivation path to recover pre-ledger-live assets
+APP_LOAD_PARAMS += --path "44'/5718350'" --path "44'/60'"
 DEFINES += CHAINID_UPCASE=\"WAN\" CHAINID_COINNAME=\"WAN\" CHAIN_KIND=CHAIN_KIND_WANCHAIN CHAIN_ID=1
 APPNAME = "Wanchain"
 else ifeq ($(CHAIN),kusd)


FabriceDautriat:
Wanchain and Ethereum are using two different coin types that are officially registered in slip44.
Wanchain’s coin type is: 5718350
Ethereum’s coin type is: 60

As such, the derivation paths are different, leading to different addresses:
m/44'/5718350'/0' is the correct path for Wanchain
m/44/60'/0' is the correct path for Ethereum

**For security reasons, Ledger devices do not allow access to another derivation path than the one belonging to the application. Thus, the Ethereum path is not accessible in the Wanchain application.**

To be consistent, when you are transferring WAN tokens to your own account, you should only send to addresses given by the Wanchain application, not the Ethereum application.

It's possible that the Wanchain application wrongly used the Ethereum path in the past or that WAN tokens were sent to an Ethereum address by mistake. 

In any case, you should still be able to access your tokens by entering your 24-word recovery phrase on the Wanchain wallet and by manually selecting the Ethereum derivation path.

_Please note that this procedure may compromise the security offered by your Ledger hardware wallet. Ledger does not accept the responsibility for any outcome this procedure may have. 

If you do proceed, please send all other crypto assets to addresses generated from a new recovery phrase. Please refer [to this article](https://support.ledger.com/hc/en-us/articles/115005297709-Export-your-accounts) for more information._

tyrion70:
Hi Fabrice,

As indicated in the PR, this "feature" of Ledger to not allow other paths was added after the fact. The old Ledger application did allow access to the Ethereum Derivation path and Wanchain used to be registered with the Ethereum derivation path (Just like POA for example which btw still runs under the Eth derivation path even though they are registered now with their own (178).

I hope you are not suggesting that people should compromise the security of their Ledger wallet by entering the passphrase of that same wallet anywhere else than in their ledger.. 

I hope you will still merge this PR because I can't support people with the suggestion to enter their passphrase anywhere outside the Ledger, especially since you disabled the access to that derivation path without prior notice to those users. Maybe we could meet in the middle by allowing it for a certain timeframe.

If not I would like to ask how you would for example migrate POA in te future to their own derivation path? All I'm asking is for the same migration path.

Thanks again,
Peter

FabriceDautriat:
Dear Peter,

Thanks for noticing the Ethereum path is still available for POA. 
You are right about this and I’ll try to explain the difference.

When dealing with blockchain forking or deriving from Ethereum, there are two scenario to consider:

1. The new chain has its own ChainID (following EIP155).
2. Thenew chain still uses Ethereum ChainID or the same ChainID as another chain.

In the first case, accessing Ethereum derivation path will indeed allow the user to see his Ethereum address when using a compatible wallet application _(using MyEtherWallet and selecting POA network for instance)._
But signing a transaction from this (Ethereum) address won’t make it valid because the ChainID won’t match. Therefore it’s possible for users to wrongly send funds to the Ethereum address and then have them “locked” there.

In the second case, accessing Ethereum derivation path will not only allow to see the address but also sign transaction from this address. 
In this case, even though the funds are absolutely retrievable, the issue is different : we cannot guarantee which funds are being transferred. For all we know, the wallet application (for instance https://mywanwallet.io/) could be tricking the user letting him know he’s signing a WAN transaction when in reality it’s an Ethereum transaction. This is the kind of security issue we can’t allow.

In the first case, yes, some funds might be difficult to access. But users can still retrieve them following the export procedure mentioned in my last message.
In the second case, our only option is to prevent Wanchain application to access any other derivation paths than the one it’s been registered on.

In any cases, you are perfectly correct, the [export your account ](https://support.ledger.com/hc/en-us/articles/115005297709-Export-your-accounts) procedure is not a perfect option to retrieve funds that could be locked.

We are currently working on a “ETH-Network” app that will allow users to retrieve those funds without exposing their 24 words. Please note that even if it improves the security, it’s still far from perfect, and it won’t solve the risk mentioned above where one MEW-like wallet would try to trick the user.

The only good solution is to lock the paths for coins without unique EIP155 ChainID.

tyrion70:
Hi @FabriceDautriat,

Thanks for your extensive answer, however the risk you are describing does not apply to Wanchain and I'll explain why :)

Wanchain uses the same ChainId as Ethereum and in that way does not follow EIP155, but the transaction format of Wanchain transactions are unique in the sense that an extra field has been added especially for Wan. This field (TxType), although is mandatory on the wanchain transactions so an Ethereum transaction could never succeed on Wanchain and vice versa. 

In hindsight I wish the ChainId was different because if would have made discussions like this a lot easier, but you see, there is a way for you to determine a transaction is for Wan or Eth so therefor you guarantee it. There is even a special line of code to enforce this:
https://github.com/LedgerHQ/ledger-app-eth/blob/565e158d0ad021faf777e87e33740420b58d03b9/src/main.c#L2353

So again, I would really appreciate this PR to be merged so people have a safe way of accessing their coins again :)

Thanks,
Peter

FabriceDautriat:
@TamtamHero can you please review the line of code mentioned above ?

@btchip : Can we get your feedback on this ?
- If chainID != 1 ==> OK to use 44'/60' ?
- If chainID = 1 ==> Is there a rule we can define to determine if we allow 44'/60' ?

In any case, we should update our guidelines and documentation to reflect this decision.

Thanks !

PR #95, opened 2020-04-14
Removed volatile qualifiers in many place

kingofpayne:
There is no justification here in declaring variables as volatile.


btchip:
Rebased on Starkware branch to be merged -> closed

PR #96, opened 2020-04-14
Minor code improvement

kingofpayne:
Using MIN instead of ternary operator for better code in ethUstream.c.


kingofpayne:
When compiling the binary remains the same.

PR #118, opened 2020-12-03
Add setExternalPlugin command

TamtamHero:
Add a new command to set the name of the plugin that will be used to parsed contract data in the next transaction.
This command fails early if the plugin is not installed on the device.


pscott:
I think this one has been "silently" merged on master by another branch (now that plugins are supported). Do you confirm we can close this one @TamtamHero ?

PR #126, opened 2021-02-05
Add setting display nonce

pscott:
This PR aims at adding a special setting to be able to display the nonce when approving a transaction.
The setting is off by default.

Additionally, this PR reworks how the UX_FLOW array was constructed, making it dynamic (this will help if we wish to add additional settings).

### DIFF
--- a/src_features/signTx/logic_signTx.c
+++ b/src_features/signTx/logic_signTx.c
@@ -229,7 +229,7 @@ void computeFees(char *displayBuffer, uint32_t displayBufferSize) {
 }
 
 void finalizeParsing(bool direct) {
-    char displayBuffer[50];
+    char displayBuffer[79]; // required to store the string representation of uint256_t max
     uint8_t decimals = WEI_TO_ETHER;
     uint8_t *ticker = (uint8_t *) PIC(chainConfig->coinName);
     ethPluginFinalize_t pluginFinalize;
@@ -352,6 +352,15 @@ void finalizeParsing(bool direct) {
                        displayBuffer,
                        sizeof(displayBuffer));
         compareOrCopy(strings.common.fullAmount, displayBuffer, called_from_swap);
+    }
+    // Prepare nonce to display
+    if (genericUI) {
+        uint256_t nonce;
+        convertUint256BE(tmpContent.txContent.nonce.value,
+                         tmpContent.txContent.nonce.length,
+                         &nonce);
+        tostring256(&nonce, 10, displayBuffer, sizeof(displayBuffer));
+        compareOrCopy(strings.common.nonce, displayBuffer, called_from_swap);
     }
     // Compute maximum fee
     if (genericUI) {
--- a/src_features/signTx/logic_signTx.c
+++ b/src_features/signTx/logic_signTx.c
@@ -229,7 +229,7 @@ void computeFees(char *displayBuffer, uint32_t displayBufferSize) {
 }
 
 void finalizeParsing(bool direct) {
-    char displayBuffer[79]; // required to store the string representation of uint256_t max
+    char displayBuffer[79];  // required to store the string representation of uint256_t max
     uint8_t decimals = WEI_TO_ETHER;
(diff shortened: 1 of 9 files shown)


src_features/signTx/logic_signTx.c
```diff
@@ -371,11 +380,7 @@ void finalizeParsing(bool direct) {
         io_seproxyhal_touch_tx_ok(NULL);
     } else {
         if (genericUI) {
-            ux_flow_init(
-                0,
-                ((dataPresent && !N_storage.contractDetails) ? ux_approval_tx_data_warning_flow
-                                                             : ux_approval_tx_flow),
-                NULL);
+            ux_approve_tx(dataPresent);
```
pscott, line 383: Either we pass the bool `dataPresent`, or we could store that in a global.

src_features/signTx/ui_flow_signTx.c
```diff
@@ -140,19 +148,23 @@ UX_STEP_NOCB(ux_approval_tx_data_warning_step,
...
-        &ux_approval_tx_3_step,
-        &ux_approval_tx_4_step,
-        &ux_approval_tx_5_step,
-        &ux_approval_tx_6_step);
+const ux_flow_step_t *ux_approval_tx_flow_[9];
+
+void ux_approve_tx(bool dataPresent) {
+    int step = 0;
+    ux_approval_tx_flow_[step++] = &ux_approval_tx_1_step;
+    if (dataPresent && !N_storage.contractDetails) {
```
pscott, line 156: Notice the `!N_storage.contractDetails`. I copied it from https://github.com/LedgerHQ/app-ethereum/blob/734b200ba16775517da1fd90808a30ee880d10c4/src_features/signTx/logic_signTx.c#L376-L377 but I found it hard to understand.
My initial guess would've been `if (dataPresent && N_storage.contractDetails)`...

TamtamHero: `contractDetails` is a user setting
`contractDetails` == 1 -> we display all the variables in the contract data
`contractDetails` == 0 -> we just show a warning indicating that the transaction contains some contract interaction

src_features/signTx/logic_signTx.c
```diff
@@ -229,7 +229,7 @@ void computeFees(char *displayBuffer, uint32_t displayBufferSize) {
 }
 
 void finalizeParsing(bool direct) {
-    char displayBuffer[50];
+    char displayBuffer[79];  // required to store the string representation of uint256_t max
```
TamtamHero, line 232: let's take a more memory friendly approach: there's 112 million ETH, let's assume it inflates to 1 billion to have a realistic worst case scenario: the biggest ETH amount one could potentially spend has a length of 10 digits + '.' + 18 decimals digits + '\0' == 30 bytes

src_features/signTx/logic_signTx.c
TamtamHero: It looks good to me this way :+1:

src_features/signTx/ui_flow_signTx.c
pscott: Ohhhhh yeah haha I thought `display_warning` was the thing that would ACTUALLY display the variables. My bad !

src_features/signTx/logic_signTx.c
pscott: Hmm so that stands for the `AMOUNT` field, but not for the `nonce` field. One could create an _almost_ infinite amount of tx (this might if Ethereum gets to a point where TPS is very high and there are connected device spamming the network...).
Admittedly though, 79 would mean that one has to do A LOT of transactions hahaha. I'm fine changing this back to 50, just wanted to make sure that this was OK with the reviewers :)

TamtamHero: I'm sure we'll boil the oceans before someone reaches a nonce with 50 digits, we can safely go back to 50 bytes

PR #127, opened 2021-03-25
Fix compilation warnings for 1.6.1-2 and NanoX sdk

pscott:


Makefile
```diff
@@ -290,7 +290,7 @@ endif
 CC       := $(CLANGPATH)clang
 
 #CFLAGS   += -O0
-CFLAGS   += -O3 -Os -I/usr/include
+CFLAGS   += -O3 -Os -I/usr/include -Wno-format-invalid-specifier -Wno-format-extra-args -Wno-main
```
pscott, line 293: Had to add those nasty flags at compilation because `H` flag isn't recognised as a valid specifier by `snprintf` ^^

PR #131, opened 2021-04-15
Support EIP2718 (transaction types) and EIP2930 (access list transactions) and display ChainID

pscott:
[EIP 2718](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2718.md) introduces Typed Transaction. Transactions can now have a single byte between 0 and 0x7f to uniquely identify their type.

This PR adapts the txdata parser to support those types of transactions.
It also adds support for EIP2930 which adds a new transaction type (`txtype 01`).
This PR also adds the feature of displaying the ChainID when it's different than 1.

### DIFF
--- a/doc/ethapp.asc
+++ b/doc/ethapp.asc
@@ -474,6 +474,7 @@ The following standard Status Words are returned for all APDUs - some specific S
 |===============================================================================================
 | *SW*     | *Description*
 |   6501   | TransactionType not supported
+|   6502   | Not enough space for conversion from uint32_t to string
 |   6700   | Incorrect length
 |   6982   | Security status not satisfied (Canceled by user)
 |   6A80   | Invalid data
--- a/doc/ethapp.asc
+++ b/doc/ethapp.asc
@@ -474,7 +474,7 @@ The following standard Status Words are returned for all APDUs - some specific S
 |===============================================================================================
 | *SW*     | *Description*
 |   6501   | TransactionType not supported
-|   6502   | Not enough space for conversion from uint32_t to string
+|   6502   | Output buffer too small for snprintf input
 |   6700   | Incorrect length
 |   6982   | Security status not satisfied (Canceled by user)
 |   6A80   | Invalid data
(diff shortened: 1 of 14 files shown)


doc/ethapp.asc
```diff
@@ -473,6 +473,8 @@ The following standard Status Words are returned for all APDUs - some specific S
 [width="80%"]
 |===============================================================================================
 | *SW*     | *Description*
+|   6501   | TransactionType not supported
+|   6502   | Not enough space for conversion from uint32_t to string
```
TamtamHero, line 477: that's very accurate yet not so clear
maybe "output buffer too small for u32_to_str input" instead?

src/utils.h
```diff
@@ -28,7 +28,12 @@ void convertUint256BE(uint8_t* data, uint32_t length, uint256_t* target);
 
 int local_strchr(char* string, char ch);
 
-uint32_t getV(txContent_t* txContent);
+// `itoa` for uint32_t.
+void u32_to_str(char* dest, uint8_t dest_size, uint32_t in);
```
TamtamHero, line 32: Prototype differs from definition :robot: 
I prefer the prototype btw :angel:

src/utils.c
```diff
@@ -53,22 +53,62 @@ int local_strchr(char *string, char ch) {
...
+    } else if (strict && size != 0) {
+        PRINTF("Unexpected format\n");
         THROW(EXCEPTION);
     }
-    return v;
+    return res;
+}
+
+// Converts a uint32_t to a string.
+void u32_to_str(char *dest, uint32_t in, uint8_t dest_size) {
```
TamtamHero, line 76: actually you can use `snprintf` to do this :f:
`snprintf(dest, dest_size, "%d ", in);`

TamtamHero (changes requested):

src/utils.h
pscott: Removed for snprintf :)

PR #133, opened 2021-04-22
Display Validator address, add bound to withdrawal derivation index, check deposit contract addresss.

pscott:
This PR adds three new features:
1. Display the validator address on screen when depositing.
2. Abort signing when the account index of the withdrawal  key is higher than `INDEX_MAX`.
3. Check that the `destination` field of the transaction is EXACTLY the same as the deposit contract address (sources for the address: [1](https://docs.ethhub.io/ethereum-roadmap/ethereum-2.0/deposit-contract/), [2](https://etherscan.io/address/0x00000000219ab540356cbb839cbe05303d7705fa), [3](https://ethereum.org/en/eth2/deposit-contract/)).


src_features/setEth2WithdrawalIndex/withdrawal_index.h
```diff
@@ -1,5 +1,7 @@
 #include "stdint.h"
 
+#define INDEX_MAX 1337  // scott
```
pscott, line 3: What value do we want to put in here ? 100 ? 1000 ?

src_features/setEth2WithdrawalIndex/cmd_setEth2WithdrawalIndex.c
```diff
@@ -18,6 +19,9 @@ void handleSetEth2WithdrawalIndex(uint8_t p1,
     }
 
     eth2WithdrawalIndex = U4BE(dataBuffer, 0);
+    if (eth2WithdrawalIndex > INDEX_MAX) {
+        THROW(0x6A80);  // scott throw this error code or create new one ?
```
pscott, line 23: My comment is to be addressed.... should we create a new error code ?

TamtamHero (changes requested):

TamtamHero (approved):

PR #151, opened 2021-05-28
Add comment about EIP-155

pscott:
This particular piece of code was not commented and I think we needed to document this somewhere :)


pscott:
See #157

PR #152, opened 2021-05-28
Add network display

### LINKED ISSUE #5: ETC confirmation show ETH
Hi, when confirming a ETC transaction, the recap showed on the Nano's screen show it like it is a ETH transaction, ("Amount ETH 0.5" instead of "Amount ETC 0.5").

pscott:
Display the name of the network rather than the Chain ID.
A simple chain ID is rather confusing to most users, so this PR should improve UX :)

Closes #5


TamtamHero (approved):

pscott:
We need to check that other eth clones don't have networks displayed 24/7...

PR #157, opened 2021-06-11
Support EIP1559

### LINKED ISSUE #150: EIP-1559 Support
Greetings.

I am working on EIP-1559 support at MetaMask and am doing discovery & identifying unknowns. When we pass a transaction to hw-app-ethereum for signing we send it a serialized transaction. We use `@ethereumjs/tx` to generate the serialized transaction which will take into consideration all of the rlp encoding needed for the new fields. When we get the `v, r, s` values back from ledger we append them to the transaction. 

What work is required by Ledger to be able to support EIP-1559 and is there a tracking issue? I assume that ledger unserializes the transaction to display the data to the user on the hardware wallet. A similar question could be asked about EIP-2930 and any future transaction type added under the EIP-2718 Typed transaction envelope.

pscott:
This PR adds support for [`EIP1559` style transactions](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md).

It provides base support for those transactions.

Additionally, it removes the `data Present` warning in case the data is set to `0` (in which case there is no data in the transaction...). This is a small UX improvement :)

Closes #150


src/ui_flow.c
```diff
@@ -113,10 +123,21 @@ UX_STEP_CB(
...
     });
 
+  UX_STEP_CB(
+    ux_settings_flow_4_step,
+    bnnn,
+    switch_settings_display_fee_details(),
+    {
+      "Fee Details",
+      "Display fee details",
+      "when possible",
```
pscott, line 133: I added "when possible" because when dealing with non EIP1559 transactions, there are no details. But I'm wondering if it wouldn't be better to just remove this sentence.

brad-decker:
Hey @pscott thanks for taking this on! If I can be of service let me know. Not very fluent with the language but I might be able to drum up community support if you want it.

pscott:
> Hey @pscott thanks for taking this on! If I can be of service let me know. Not very fluent with the language but I might be able to drum up community support if you want it.

Hey! :)
The work is done and is now under review by @TamtamHero :D Will add some tests cases in the future but I don't think the support of EIP-1559 will need much more code for Ledger :)
Yayyy London :D

src/ui_flow.c
TamtamHero: "when available" maybe ?

src/shared_context.h
```diff
@@ -165,7 +166,8 @@ typedef enum {
 typedef struct txStringProperties_t {
     char fullAddress[43];
     char fullAmount[50];
-    char maxFee[50];
+    char maxFee[50]; // Used as BaseFee when detailing fees
+    char priorityFee[50];
```
TamtamHero, line 170: oh my god 50 bytes :scream:

pscott: F

src/ui_flow.c
pscott: d724d4196ccd34b3fbba5d585cbc21c1706fff01

pscott:
TODO: add a `HAVE_DETAILED_FEES` that would force the setting to be set, compile it, and add a test to make sure that the detailed fees are correctly displayed :D

examples/signTx.py
```diff
@@ -91,6 +96,8 @@ def parse_bip32_path(path):
 )
 
 encodedTx = encode(tx, UnsignedTransaction)
+# encodedTx = bytearray.fromhex(
```
unparalleled-js, line 99: Wish there was a comment here saying that this the EIP-1559
(I am pretty sure this is the EIP-1559!)

pscott: Indeed this was a test for an EIP1559 tx (similar to the test for an EIP2930 tx at line 102). Feel free to submit a PR to add a comment :D @unparalleled-js :)

unparalleled-js: Do you happen to have any advice for encoding a transaction this way in python?
I am struggling with that part

pscott: I do not :) I probably used web3js or something like that to encode the transaction :)

PR #165, opened 2021-06-23
feat: add yearn deposit and withdraw methods

steffenix:


pscott:
@steffenix Will have a look at this. We already have a couple of ongoing PRs and work to do, but will review this in the future! :D

steffenix:
If you have any ideas on why it seems like selectors are not detected so that I can continue, that would be great

pscott:
> If you have any ideas on why it seems like selectors are not detected so that I can continue, that would be great

Yup it's because you need to try those with a local version of `ledgerjs` and add your selectors to [the list of supported dapps](https://github.com/LedgerHQ/ledgerjs/blob/master/packages/cryptoassets/data/dapps/ethereum.js).
You won't be able to try it out with Metamask (as it doesn't use your local version of ledgerjs) so you'll need to craft transactions yourself. Have a look at the `tests` folder inside the ethereum-app, you should be able to add tests for your work :D

steffenix:
Thanks,

I had a look at the file and it seems I need to specify a specific contract address, at yearn we have about 50 different contracts that we update on a regular basis making it close to impossible to keep updating the address list on every update.

steffenix:
@pscott I am looking at https://github.com/LedgerHQ/ledgerjs/blob/1e6190ff89b7b3f0ae852719410eb3a33e08c43f/packages/cryptoassets/data/dapps/ethereum.js#L6 what does serialized_data & signature stands for?

pscott:
> @pscott I am looking at https://github.com/LedgerHQ/ledgerjs/blob/1e6190ff89b7b3f0ae852719410eb3a33e08c43f/packages/cryptoassets/data/dapps/ethereum.js#L6 what does `serialized_data` & `signature` stands for?
> 
> I see it used here
> setExternalPlugin(this.transport, payload, signature);
> Would like to know how to generate it.
> 
> serialized_data: 08436f6d706f756e6 <> contract address <> selector
> Where is 08436f6d706f756e6 from?
> How do I sign that?

Hi,
The actual crypto asset list repository is still private but will be going public soon. The signatures for this specific package are signatures of Ledger. In order to test, the thing you need to do is deactivate the signature check inside the ethereum-app, so that it will work with your "in-house" signatures

steffenix:
ok, 

What kind of signature is that?
Where does `08436f6d706f756e6` comes from?

pscott:
> ok,
> 
> What kind of signature is that?
> Where does `08436f6d706f756e6` comes from?

@TamtamHero can provide more information on that matter :)

TamtamHero:
The serialization of this command is documented here: https://github.com/LedgerHQ/app-ethereum/blob/master/doc/ethapp.asc#set-external-plugin
Only ledger can sign these information, so you have to bypass the signature check when developing. To do so, comment out these lines: https://github.com/LedgerHQ/app-ethereum/blob/master/src_features/setExternalPlugin/cmd_setExternalPlugin.c#L33-L47
And simply skip sending the signature at the end of the command. 

Sorry, it's a bit rough around the edges, but we're working on it

steffenix:
So 08436f6d706f756e6 should be and len(pluginName) || pluginName.
How do I get from `yearn_plugin` to a hexadecimal value?

steffenix:
@TamtamHero the doc you are pointing me to is for an external plugin, we have over 50 different contracts so we need to use an internal plugin as I can see from the doc.

```
Creating an internal plugin

Internal plugins are triggered on specific selectors. You can modify src/eth_plugin_internal.c to add your mapping.

Other specific mappings can be also added by modifying the common dispatcher
Creating an external plugin

An external plugin is a library application named after the base64 encoding of the 20 bytes smart contract address
```

PR #169, opened 2021-07-02
Fix additional screens initialization

pscott:
`additional_screens` was introduced in a PR previously but isn't properly initialized in some cases.

### DIFF
--- a/src/eth_plugin_handler.c
+++ b/src/eth_plugin_handler.c
@@ -225,9 +225,7 @@ eth_plugin_result_t eth_plugin_call(int method, void *parameter) {
         params[2] = (uint32_t) parameter;
         BEGIN_TRY {
             TRY {
-                PRINTF("Jumping to lib\n");
                 os_lib_call(params);
-                PRINTF("Back from lib\n");
             }
             CATCH_OTHER(e) {
                 PRINTF("Plugin call exception for %s\n", alias);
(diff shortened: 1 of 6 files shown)


src/eth_plugin_handler.c
```diff
@@ -21,12 +21,8 @@ void eth_plugin_prepare_finalize(ethPluginFinalize_t *finalize) {
     memset((uint8_t *) finalize, 0, sizeof(ethPluginFinalize_t));
 }
 
-void eth_plugin_prepare_provide_token(ethPluginProvideToken_t *provideToken,
-                                      tokenDefinition_t *token1,
-                                      tokenDefinition_t *token2) {
+void eth_plugin_prepare_provide_token(ethPluginProvideToken_t *provideToken) {
     memset((uint8_t *) provideToken, 0, sizeof(ethPluginProvideToken_t));
-    provideToken->token1 = token1;
-    provideToken->token2 = token2;
```
pscott, line 29: eth_plugin_prepare_provide_token is now called BEFORE getting token information. Hence the safe removal of those fields

src/eth_plugin_handler.c
```diff
@@ -128,7 +132,6 @@ eth_plugin_result_t eth_plugin_perform_init(uint8_t *contractAddress,
 eth_plugin_result_t eth_plugin_call(int method, void *parameter) {
     ethPluginSharedRW_t pluginRW;
     ethPluginSharedRO_t pluginRO;
-    char tmp[PLUGIN_ID_LENGTH];
```
pscott, line 131: This was unused

pscott:
See #170

PR #181, opened 2021-08-06
Remove check of v length

pscott:
Remove check for v length in order to support bigger chain IDs


src_features/signTx/logic_signTx.c
```diff
@@ -257,22 +257,6 @@ void prepareFeeDisplay() {
                        sizeof(strings.common.maxFee));
 }
 
-uint32_t get_chainID() {
```
pscott, line 260: this fn was not used... there's a second `get_chainid()` that is actually used ^^

PR #187, opened 2021-09-09
WIP: Add moonrriver

joelamouche:


Makefile
```diff
@@ -179,6 +179,10 @@ else ifeq ($(CHAIN),tomochain)
 APP_LOAD_PARAMS += --path "44'/889'"
 DEFINES += CHAINID_UPCASE=\"TOMOCHAIN\" CHAINID_COINNAME=\"TOMO\" CHAIN_KIND=CHAIN_KIND_TOMOCHAIN CHAIN_ID=88
 APPNAME = "TomoChain"
+else ifeq ($(CHAIN),moonriver)
+APP_LOAD_PARAMS += --path "44'/60'"
```
jibeee, line 183: Derivation path should be 44'/1285', and not 44'/60'.

jibeee (commented):
Please change derivation path.

Makefile
joelamouche: We want the same derivation path as ethereum. Otherwise derived addresses from other ethereun wallets wont have the same address

joelamouche: https://github.com/LedgerHQ/app-ethereum/pull/215/files

PR #193, opened 2021-09-17
Add a UI warning when contract data are not enabled in the settings

TamtamHero:
Contract data setting is disabled by default, and updating the Ethereum app resets it to its default value.
This UI warning will remind users that they have to enable contract data if they try to sign a transaction which contains some.


Makefile
```diff
@@ -285,7 +285,7 @@ endif
 
 
 # Enabling debug PRINTF
-DEBUG:=0
+DEBUG:=1
```
pscott, line 288: :eyes:

tests/build_local_test_elfs.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 # FILL THESE WITH YOUR OWN SDKs PATHS
-NANOS_SDK=$TWO
-NANOX_SDK=$X
+# NANOS_SDK=
+# NANOX_SDK=
```
pscott, line 5: Sank you

tests/src/test.fixture.js
```diff
@@ -0,0 +1,75 @@
...
+        let zemu_args;
+        let sim_options;
+        if(device === "nanos"){
+            zemu_args = [NANOS_ELF_PATH];
+            sim_options = sim_options_nanos;
+        }
+        else{
+            zemu_args = [NANOX_ELF_PATH];
+            sim_options = sim_options_nanox;
+        }
```
pscott, line 51: Nothing important but weird spacing here.
Would go for
```js
if (device === "nanos") {
//
}
else {
//
}
```

tests/src/contract_data_warning.test.js
```diff
@@ -0,0 +1,38 @@
...
+import { expect } from "../jest";
+import { TransportStatusError } from "@ledgerhq/errors";
+import { waitForAppScreen, zemu } from './test.fixture';
+import Zemu from '@zondax/zemu';
+
+test('[Nano S] Try to blind sign with setting disabled', zemu("nanos", async (sim, eth) => {
+  // disable blind signing
+  await sim.navigateAndCompareSnapshots('.', 'nanos_disable_blind_signing', [2, 0, 0, 3, 0]);
+
+  // we can't use eth.signTransaction because it detects that contract data is disabled and fails early
```
pscott, line 12: Interesting. Thanks for pointing this out in the comments! :)

pscott (approved):
Minor changes could be made but we can merge as is if you prefer @TamtamHero :D

Makefile
TamtamHero: Oof

tests/src/contract_data_warning.test.js
TamtamHero: :pet-myself:

pscott:
@TamtamHero Looks like tests are failing :D

PR #202, opened 2021-10-05
feat: goerli build

Neozaru:
- Adds `goerli` as target chain for build, with same profile as `ropsten`
- Adds 4 extra DeversiFi test ERC20 tokens


src/tokens.c
```diff
@@ -37,6 +38,23 @@ const tokenDefinition_t const TOKENS_EXTRA[NUM_TOKENS_EXTRA] = {
...
+     "tBTC ",
+     18},
+    {{0xa8, 0xf3, 0x14, 0x4f, 0xea, 0x2c, 0x37, 0x5a, 0xd0, 0x58,
+      0xec, 0x12, 0x09, 0x9a, 0x5a, 0x21, 0xa2, 0x6f, 0xe9, 0x96},
+     "tDVF ",
+     18},
+    {{0x4c, 0xda, 0xbe, 0xc1, 0x2a, 0x39, 0x7f, 0xb6, 0xef, 0xaf,
+      0x46, 0x13, 0xd5, 0xdf, 0xd7, 0x9b, 0x30, 0x9a, 0xe9, 0xfa},
+     "tXDVF ",
+     18},
```
FabriceDautriat, line 57: As far as I know this token.c file will not be used. 
Instead we should add your token to the list of ERC20 supported tokens. (cc @TamtamHero for confirmation)

Can you please confirm these same contract don't already exist on Etheruem mainnet ?

Neozaru: Adding those tokens here + adding `HAVE_TOKENS_EXTRA_LIST` for `goerli` in the Makefile did work for our cases (the goal is for us to be able to call `provideERC20TokenInformation` with `00{tokenAddress}...{chainId}`.
Before adding those tokens in that file, the above won't work. After adding them I've got the behavior I expected.
The tokens in the list have been freshly minted on Goerli with fresh smart-contract addresses. It doesn't collude with anything on mainnet.

TamtamHero: @FabriceDautriat  It works because Goerli compiles as a standalone Eth app, unlike clones (cf `DEFINES_LIB=` in the Makefile, same for ropsten)

TamtamHero:
Can you please sign your commits @Neozaru ? This is a requirement on this repo.
https://stackoverflow.com/questions/13504983/retroactively-sign-git-commits

TamtamHero:
it's all good, thanks !

PR #228, opened 2021-12-16
change dataContext.tokenContext.fieldIndex to uint16_t 

Benjyskan:
...to avoid plugin's 'context->parameterOffset' overflow on large tx

### DIFF
--- a/src/shared_context.h
+++ b/src/shared_context.h
@@ -53,7 +53,7 @@ typedef struct tokenContext_t {
     uint8_t pluginStatus;
 
     uint8_t data[INT256_LENGTH];
-    uint8_t fieldIndex;
+    uint16_t fieldIndex;
     uint8_t fieldOffset;
 
     uint8_t pluginUiMaxItems;


adrienlacombe-ledger:
exemple of overflow https://etherscan.io/tx/0xf4428462060cc443e9d19a5048ad31dcae46497e1133a0f3016e31c62596067e

apaillier-ledger:
Hello @Benjyskan, thank you for the PR !  :+1: 
I reproduced it with an ERC-1155 batch transfer across more than 255 NFT IDs, `fieldIndex` is indeed two small with 8 bits.
* Could you change it to 16 instead of 32 bits though ? 32 becomes so large it's a bit unrealistic and even if it could be this large, then we would be in trouble with `parameterOffset` being only 32 bits and could overflow if `fieldIndex` is too large [here](https://github.com/LedgerHQ/app-ethereum/blob/1.10.1/src_features/signTx/logic_signTx.c#L130).
* Could you please rebase your PR with develop, so that it is up to date and triggers the CI ?

apaillier-ledger (approved):

PR #231, opened 2022-01-14
[fix] Get printable amount handle should not parse configuration for …

lpascal-ledger:
…ticker and decimal if the amount is a fee

External such as exchange could use an Ethereum handle to get a printable view of ETH amount or fees (using the `get_printable_amount_parameters_t` structure).

In case of fees, the Ethereum would currently return an error if the handle is trigger without indication in `get_printable_amount_parameters_t:coin_configuration` (or `get_printable_amount_parameters_t:coin_configuration_length == 0`) even if these fields are eventually not used.

This PR fixes this behavior.


apaillier-ledger (approved):
OK

TamtamHero (approved):

PR #247, opened 2022-02-10
Fix NFT transaction crash

apaillier-ledger:
* Fix crash when the PROVIDE_NFT_INFORMATION APDU was not received
* Fix screen number for the ERC-1155 setApprovalForAll function
* Removed unused screens from ERC-721 & 1155 functions
* Add Zemu tests for error cases


src_plugins/erc1155/erc1155_ui.c
```diff
@@ -26,7 +28,10 @@ static void set_approval_for_all_ui(ethQueryContractUI_t *msg, erc1155_context_t
             break;
         case 2:
             strlcpy(msg->title, "NFT Address", msg->titleLength);
-            getEthDisplayableAddress(msg->pluginSharedRO->txContent->destination,
+            // In case of no PROVIDE_NFT_INFO, we already have the address from the SET_PLUGIN
+            nft_addr = ((msg->item1) ? ((uint8_t *) msg->item1->nft.contractAddress)
```
lpascal-ledger, line 32: This operation is not that complicated, but repeated 6 times, and with the same comment (the fastest self-outdating thing in programming) also repeated 6 times, it could benefit from being factorized in a dedicated function `const uint8_t *get_NFT_address(ethQueryContractUI_t const * const msg)`.

greenknot (approved):

src_plugins/erc1155/erc1155_ui.c
apaillier-ledger: Done in ce2df92741a0

PR #265, opened 2022-02-14
Add Conflux eSpace (chain ID 1030)

Thegaram:
This PR adds a new chain (eth clone) for the Conflux eSpace, a fully EVM-compatible subsystem of the Conflux blockchain. Users can add Conflux eSpace to Metamask through https://chainlist.org.

(Note that there is a separate app for the original Conflux Core which is not compatible with Metamask.)


Thegaram:
Closing in favor of #282.

PR #268, opened 2022-02-21
add cronos chain supoprt

crypto-facs:
Add support for Cronos EVM chain

NOTE: I am missing the icons


apaillier-ledger:
Done in #334

PR #279, opened 2022-03-19
BugFix - add missing variable in processTx's PRINTF

Benjyskan:


apaillier-ledger:
Nice catch, thank you for your contribution.

PR #321, opened 2022-06-27
Fuzz support

jmartins-ledger:
## Description

Fix vulnerabilities found by fuzzing.
- fix multiple oob reads by refactoring bip32 parsing
- fix stack overflow on tostring256()
- fix processV() missing check which allowed overflows

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src_features/signTx/cmd_signTx.c
+++ b/src_features/signTx/cmd_signTx.c
@@ -12,42 +12,31 @@ void handleSign(uint8_t p1,
                 unsigned int *tx) {
     UNUSED(tx);
     parserStatus_e txResult;
-    uint32_t i;
 
     if (os_global_pin_is_validated() != BOLOS_UX_OK) {
         PRINTF("Device is PIN-locked");
         THROW(0x6982);
     }
     if (p1 == P1_FIRST) {
-        if (dataLength < 1) {
-            PRINTF("Invalid data\n");
-            THROW(0x6a80);
-        }
         if (appState != APP_STATE_IDLE) {
             reset_app_context();
         }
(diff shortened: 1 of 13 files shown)


jibeee (changes requested):
I suggest not using `THROW` in new code for this app.

You could rewrite `parseBip32` so that it returns NULL when it fails. Also, you could create a structure "`bip32_path_t`" with a size and  a path.

`parseBip32` would look like:

```c
uint8_t *parseBip32(const uint8_t **data, uint16_t *dataLength, bip32_path_t *path);
```

That would make the function easily testable. What do you think?

src_features/signTx/cmd_signTx.c
```diff
@@ -12,43 +12,37 @@ void handleSign(uint8_t p1,
...
-                PRINTF("Invalid data\n");
-                THROW(0x6a80);
-            }
-            tmpCtx.transactionContext.bip32Path[i] = U4BE(workBuffer, 0);
-            workBuffer += 4;
-            dataLength -= 4;
-        }
+
+        tmpCtx.transactionContext.pathLength = bip32.length;
+        memcpy(tmpCtx.transactionContext.bip32Path, bip32.path, bip32.length * sizeof(uint32_t));
```
jmartins-ledger, line 34: with the new changes we can have situations like this

jibeee: Fixed in https://github.com/LedgerHQ/app-ethereum/pull/321/commits/e45e92ef6c0ea206562c147a153e2436adaa8d4e.

PR #323, opened 2022-07-05
Add CodeQL Workflow 

mlafon-ledger:
## Description

Add analysis by CodeQL each time a push or a PR is made on the branch master & develop

## Changes include

- [ ] Tests
- [ ] Documentation
- [ X] Other (for changes that might not fit in any category)


.github/workflows/codeql-workflow.yml
```diff
@@ -0,0 +1,42 @@
...
+      - develop
+  pull_request:
+    branches:
+      - master
+      - develop
+    paths-ignore:
+      - '.github/workflows/*.yml'
+
+jobs:
+  job_build_debug_nano_s:
```
jibeee, line 16: You should build app for every device. Moreover, job name is misleading. You should keep the default names proposed when setting up the workflow:

```yaml
jobs:
  analyze:
    name: Analyze
```

.github/workflows/codeql-workflow.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v2
+        with:
+          languages: ${{ matrix.language }}
+          queries: security-and-quality
+
+      - name: Build
+        run: |
+          make clean
+          make DEBUG=1 ALLOW_DATA=1
```
jibeee, line 38: `ALLOW_DATA=1` is not necessary here. It is used to automatically enable an option in the app for testing purposes.

Also, I think you shouldn't compile app in debug mode: code scanning is more useful for the release version.

jibeee (changes requested):

.github/workflows/codeql-workflow.yml
mlafon-ledger: Done

.github/workflows/codeql-workflow.yml
mlafon-ledger: Done

jibeee (approved):
Good PR.

PR #339, opened 2022-09-12
EIP-191 UI improvements

apaillier-ledger:
## Description

The current EIP-191 implementation only shows the first 255 bytes for LNX & LNS+ and 99 for LNS of a message.
This PR aims to improve this by allowing the user to see the whole message by chunks of the same previously mentioned size with the ability to skip the rest after each chunk.
This stays 100% compatible with the currently implemented APDU.

## Changes include

- [ ] Tests
- [ ] Documentation


lpascal-ledger (approved):

src_features/signMessage/cmd_signMessage.c
```diff
@@ -1,199 +1,292 @@
...
+                               uint8_t p2,
+                               const uint8_t *const payload,
+                               uint8_t length) {
+    const uint8_t *data = payload;
 
-    feed_value_str(workBuffer, dataLength, is_value_str_ascii());
+    (void) p2;
+    processed_size = 0;
+    if (p1 == P1_FIRST) {
+        if ((data = first_apdu_data(data, (uint16_t *) &length)) == NULL) {
```
yhql, line 224: `(uint16_t *) &length` has the same undefined behaviour as first comment in this file

src_features/signMessage/cmd_signMessage.c
```diff
@@ -1,199 +1,292 @@
...
+ * @return size of data in bytes
  */
-static void init_value_str(bool is_ascii) {
-    if (is_ascii) {
-        strings.tmp.tmp[0] = '\0';  // init string as empty
-    } else {
-        strcpy(strings.tmp.tmp, "0x");  // will display the hex bytes instead
-    }
+static size_t unprocessed_length(void) {
+    return G_io_apdu_buffer[OFFSET_LC] - processed_size;
```
yhql, line 43: The data length in `G_io_apdu_buffer[OFFSET_LC]` could be a lie and not match the actual length of the data. It's kind of ok here since we're displaying everything and the user would see garbage from the previous APDU, but it can be fixed easily and for all APDUs I think, by always checking that the `rx` value returned from `io_exchange` is `G_io_apdu_buffer[OFFSET_LC] + 5`. What do you think?

yhql (changes requested):

src_features/signMessage/cmd_signMessage.c
```diff
@@ -1,199 +1,289 @@
...
+ * Send a response APDU with the given Status Word
  *
- * @param[in] c character we're checking
- * @return wether the character is special or not
+ * @param[in] sw status word
  */
-static inline bool is_char_special(char c) {
-    return ((c >= '\b') && (c <= '\r'));
+static void apdu_reply(uint16_t sw) {
+    *(uint16_t *) G_io_apdu_buffer = __builtin_bswap16(sw);
```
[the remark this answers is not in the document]
lpascal-ledger: Something like this?
```
    G_io_apdu_buffer[0] = (sw >> 8) & 0xff;
    G_io_apdu_buffer[1] = sw & 0xff;
```

src_features/signMessage/cmd_signMessage.c
lpascal-ledger: I guess I'll make a temporary `uint16_t u16_length` storing the `length` and give its pointer.

src_features/signMessage/cmd_signMessage.c
lpascal-ledger: I'm confused that this is not already checked in `main:app_main`.

yhql (dismissed):

src_features/signMessage/cmd_signMessage.c
```diff
@@ -1,199 +1,292 @@
...
-    } else if (p1 != P1_MORE) {
-        THROW(0x6B00);
+    while ((unprocessed_length() > 0) && (remaining_ui_buffer_length() > 0)) {
+        c = *(char *) unprocessed_data();
+        if (isspace(c))  // to replace all white-space characters as spaces
+        {
+            c = ' ';
+        }
+        if (isprint(c)) {
+            sprintf(remaining_ui_buffer(), "%c", (char) c);
```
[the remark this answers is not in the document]
lpascal-ledger: This may raise a bit too much code implication for me to tinker with. No sure where to keep this remark though.

src_features/signMessage/cmd_signMessage.c
lpascal-ledger: Good thing the tests catched that the length was modified then re-used further in the code.

yhql (approved):

PR #361, opened 2022-11-02
Add funcs to avoid tricking user when using a plugin

jmartins-ledger:
## Description

Usually the length of an array is sent in a parameter. Most of the times the developer simply uses `U2BE/U4BE` to get this length. It is possible to forge a tx with a `length > sizeof(uint16_t/uint32_t)` and trick the user into signing something different from what is shown.

For instance consider the following parameter:
00 ... 01 00 00 00 01

if the developer uses U2BE/U4BE, it is possible that this length is shown to the user and if it is, the user will see the length as 1.

This PR adds the following:
```
bool U2BE_from_parameter(uint8_t* parameter, uint16_t* value);
bool U4BE_from_parameter(uint8_t* parameter, uint32_t* value);
```

which besides extracting the length from the parameter also checks if the remaining bytes of a parameter are zero
## Changes include

- [ ] Tests
- [x] Documentation


doc/ethapp_plugins.asc
```diff
@@ -134,6 +134,17 @@ The following return codes are expected, any other will abort the signing proces
...
 
+There are already defined functions to extract data from a parameter:
+[source,C]
+----
+void copy_address(uint8_t* dst, const uint8_t* parameter, uint8_t dst_size);
+void copy_parameter(uint8_t* dst, const uint8_t* parameter, uint8_t dst_size);
+
+// Get the value from the beginning of the parameter (right to left) and check if the rest of it is zero
+bool U2BE_from_parameter(uint8_t* parameter, uint16_t* value);
+bool U4BE_from_parameter(uint8_t* parameter, uint32_t* value);
```
jmartins-ledger, line 145: If there are better names for these functions I will gladly change them

apaillier-ledger (approved):

PR #364, opened 2022-11-08
Fix potential crash with slow APDU transport

apaillier-ledger:
## Description

The [dummy UX_STEP](https://github.com/LedgerHQ/app-ethereum/blob/1.10.0/src_bagl/ui_flow_signMessage712.c#L39-L46) was not protected against the user clicking on it.
Over USB HID, this was not really an issue as the APDUs arrive very fast and even spamming the right button would only get the user to the "Approve" screen for a fraction of a second, resulting in a small visual glitch but nothing more.
Over a slow transport like Bluetooth the user would be able to do multiple clicks even at a normal speed, and once on the "Approve" screen clicking left would result in an instant crash of the application/device.

The solution was to entirely disable button clicks on the dummy screen while the app is waiting for more APDUs.

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -209,7 +209,7 @@ jobs:
           cd tests/ragger
           python3 -m venv ./venv
           . ./venv/bin/activate
-          pip3 install --extra-index-url https://test.pypi.org/simple/ -r requirements.txt speculos
+          pip3 install --extra-index-url https://test.pypi.org/simple/ -r requirements.txt
           # Used for the cache key
           echo "py_deps=$(pip freeze | md5sum | cut -d' ' -f1)" >> $GITHUB_ENV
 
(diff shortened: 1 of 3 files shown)


jibeee (approved):

.github/workflows/ci-workflow.yml
```diff
@@ -275,4 +275,4 @@ jobs:
         run: |
           cd tests/ragger
           . ./venv/bin/activate
-          pytest --path ./elfs --model ${{ matrix.model }}
+          pytest --path ./elfs --model ${{ matrix.model }} -s -v
```
fbeutin-ledger, line 278: `--tb=short` est bien aussi

fbeutin-ledger (approved):

.github/workflows/ci-workflow.yml
apaillier-ledger: Thanks, will look into it :+1:

PR #327, opened 2022-07-19
EIP-712 clear-signing

apaillier-ledger:
## Description

This brings EIP-712 full clear signing to the LNS+ & LNX.

Now everything is computed on the device and the app now has 3 new modes :
* The "default" view : clear signing the domain structure and the hash of the message (also computed on the device)
* The filtered way : for select contracts the device will be able to receive information from the host on which fields to show and what name to give them (to make it more user-friendly)
* The verbose way : full clear signing of both the domain and the message regardless if the host is trying to send filtering information to the app, this is done with a new in-app setting (disabled by default)

The old/existing v0 implementation that only shows the domain & message hashes that are pre-computed on the host is still supported for two reasons : 
* Since the LNS won't benefit from this PR, it needs to keep working with EIP712
* For backward compatibility on LNS+ & LNX (third party wallets like Metamask won't update to the latest [ledgerjs](https://github.com/LedgerHQ/ledger-live/tree/develop/libs/ledgerjs) instantly)

## Changes include

- [ ] Bugfix (non-breaking change that solves an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)
- [ ] Tests
- [x] Documentation
- [ ] Other (for changes that might not fit in any category)

PR #375, opened 2022-11-16
New CAL test key for CI

apaillier-ledger:
## Description

With #327, a private key was set as repository secret to be able to generate signature on-the-fly to test the CAL APDUs.
Being a secret meant that PRs from forks could not access it and Ragger tests would fail.

A new private key, that is only used for the CI tests and isn't even compiled in the production app, has been added to the repository, and now PRs coming from external developers will have functioning Ragger tests.

## Changes include

- [x] Tests
- [ ] Documentation


src/tokens.h
```diff
@@ -59,6 +66,7 @@ static const uint8_t LEDGER_SIGNATURE_PUBLIC_KEY[] = {
     0x80, 0x13, 0x73, 0x07, 0xde, 0x4d, 0xd1, 0xd4, 0x18, 0x42, 0x8c, 0xf2, 0x1a, 0x93,
     0xb3, 0x35, 0x61, 0xbb, 0x09, 0xd8, 0x8f, 0xe5, 0x79
 #endif
```
fbeutin-ledger, line 68: see above to avoid the nesting

Makefile
```diff
@@ -151,6 +151,9 @@ endif
 CAL_TESTING_KEY:=0
 ifneq ($(CAL_TESTING_KEY),0)
 DEFINES += HAVE_CAL_TESTING_KEY
+ifneq ($(CAL_TESTING_KEY),1)
+DEFINES += HAVE_CAL_TESTING_KEY_CI
+endif
 endif
```
fbeutin-ledger, line 157: Plus clair je trouve:
```
CAL_TESTING_KEY:=0
CAL_TESTING_KEY_CI:=0
ifneq ($(CAL_TESTING_KEY),0)
DEFINES += HAVE_CAL_TESTING_KEY
endif
ifneq ($(CAL_TESTING_KEY_CI),0)
DEFINES += HAVE_CAL_TESTING_KEY_CI
endif
```

Dans le code .c
#if defined (HAVE_CAL_TESTING_KEY) && defined(HAVE_CAL_TESTING_KEY_CI)
#error "CAL KEY contradiction"
#endif

tests/ragger/conftest.py
```diff
@@ -36,6 +39,13 @@ def firmware(arg_model: str) -> Firmware:
             return fw
     raise ValueError("Unknown device model \"%s\"" % (arg_model))
 
+@pytest.fixture
+def cal_key() -> Optional[SigningKey]:
```
fbeutin-ledger, line 43: Mmmh not really a fixture as it's not really a "parameter" of the test
I would suggest making a class out of it, also I suggest taking a look into the SigningAuthority defined in the exchange app tests, I instantiate and manipulate the same kind of structure. Maybe unify both and put it in ragger utils ?

src/tokens.h
```diff
@@ -51,6 +51,13 @@ static const uint8_t LEDGER_SIGNATURE_PUBLIC_KEY[] = {
     0xfe, 0x4e, 0x8c, 0xc7, 0x6b, 0x14, 0x89, 0x15, 0x0c, 0x21, 0x51, 0x4e, 0xbf, 0x44,
     0x0f, 0xf5, 0xde, 0xa5, 0x39, 0x3d, 0x83, 0xde, 0x53, 0x58, 0xcd, 0x09, 0x8f, 0xce,
     0x8f, 0xd0, 0xf8, 0x1d, 0xaa, 0x94, 0x97, 0x91, 0x83
+#else
+#ifdef HAVE_CAL_TESTING_KEY_CI
```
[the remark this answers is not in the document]
apaillier-ledger: Indeed, should be better now :+1:

Makefile
apaillier-ledger: Indeed, should be better now :+1:

tests/ragger/conftest.py
apaillier-ledger: Indeed, should be better now :+1:

fbeutin-ledger (approved):

PR #420, opened 2023-04-24
Fixes static analysis warning

apaillier-ledger:
## Description

Fixes `warning: Call to function 'strcpy' is insecure as it does not provide bounding of the memory buffer. Replace unbounded copy functions with analogous functions that support length arguments such as 'strlcpy'. CWE-119`
Introduced in 9735116f5

## Changes include

- [ ] Tests
- [ ] Documentation


src_common/uint256.c
```diff
@@ -240,7 +240,7 @@ bool tostring256(const uint256_t *const number,
 
     if (offset == outLength) {  // destination buffer too small
         if (outLength > 3) {
-            strcpy(out, "...");
+            memcpy(out, "...", 4);
```
fbeutin-ledger, line 243: strlcpy with outLength as param feels better

apaillier-ledger: Fixed in [cc283f2](https://github.com/LedgerHQ/app-ethereum/pull/420/commits/cc283f271c398088f990c24601dcdaac060eb30b)

fbeutin-ledger (approved):

PR #435, opened 2023-06-02
fix potential oob writes

jmartins-ledger:
## Description

If there is a `ticker` available check if we can write the `ticker` and the empty space.
If the `ticker_len` is larger than `out_buffer_size` then buffer is full, return failure

## Changes include

- [ ] Tests
- [ ] Documentation


cedelavergne-ledger:
Please rebase your branch also :wink:

cedelavergne-ledger:
Closing, because files have been moved/renamed.
This PR is now tracked here: https://github.com/LedgerHQ/app-ethereum/pull/566

PR #438, opened 2023-06-16
ENS on other chains

apaillier-ledger:
## Description

Now enables ENS on other Ethereum-compatible chains (with same derivation path).

## Changes include

- [x] Tests
- [ ] Documentation


src/eth_plugin_handler.c
```diff
@@ -64,7 +66,8 @@ void eth_plugin_prepare_query_contract_UI(ethQueryContractUI_t *queryContractUI,
     }
 
     queryContractUI->screenIndex = screenIndex;
-    strlcpy(queryContractUI->network_ticker, get_tx_network_ticker(), MAX_TICKER_LEN);
+    chain_id = get_tx_chain_id();
+    strlcpy(queryContractUI->network_ticker, get_displayable_ticker(&chain_id), MAX_TICKER_LEN);
```
fbeutin-ledger, line 70: What if we don't find get_displayable_ticker ?

src/eth_plugin_handler.c
```diff
@@ -64,7 +66,8 @@ void eth_plugin_prepare_query_contract_UI(ethQueryContractUI_t *queryContractUI,
     }
 
     queryContractUI->screenIndex = screenIndex;
-    strlcpy(queryContractUI->network_ticker, get_tx_network_ticker(), MAX_TICKER_LEN);
+    chain_id = get_tx_chain_id();
```
fbeutin-ledger, line 69: Is it an error if get_tx_chain_id returns 0 (not found) ?

src_common/network.c
```diff
@@ -66,73 +66,33 @@ static const network_info_t NETWORK_MAPPING[] = {
...
-
-static const network_info_t *get_network(e_net_type type) {
-    uint64_t chain_id = get_chain_id(type);
-    for (size_t i = 0; i < sizeof(NETWORK_MAPPING) / sizeof(*NETWORK_MAPPING); i++) {
-        if (NETWORK_MAPPING[i].chain_id == chain_id) {
-            return (const network_info_t *) PIC(&NETWORK_MAPPING[i]);
+static const network_info_t *get_network_from_chain_id(const uint64_t *chain_id) {
+    for (size_t i = 0; i < ARRAYLEN(NETWORK_MAPPING); i++) {
+        if (NETWORK_MAPPING[i].chain_id == *chain_id) {
+            return (const network_info_t *) &NETWORK_MAPPING[i];
```
fbeutin-ledger, line 72: PIC in this case ?

tests/ragger/requirements.txt
```diff
@@ -1,4 +1,4 @@
-ragger[speculos]>=1.7.0,<1.8.0
+ragger[speculos]>=1.8.0,<1.9.0
```
fbeutin-ledger, line 1: Why not simply latest ?

src/eth_plugin_handler.c
apaillier-ledger: `get_displayable_ticker` will always return a non-NULL ticker, in the case that `get_network_ticker_from_chain_id` returns NULL, it will return the current app/clone ticker ("ETH" for Ethereum, "MATIC" for Polygon, ...).

src/eth_plugin_handler.c
apaillier-ledger: It would be an error as that chain ID does not exist and can only be returned if it went through the `default` case of the `switch` statement in `get_tx_chain_id`.
But it has always been like this and this behavior hasn't changed, so it is out of the scope of this PR. :+1:

src_common/network.c
apaillier-ledger: `PIC` is only necessary for accessing the elements of the structs.

tests/ragger/requirements.txt
apaillier-ledger: To prevent an update with breaking changes on the implemented tests, but if this is unrealistic I could indeed change it. :+1:

fbeutin-ledger: I think we can use latest, we will never introduce breaking changes, if we really need to we'll start again from `2.0.0` anyway and it won't be a sudden update

fbeutin-ledger:
To rebase

fbeutin-ledger (approved):

PR #447, opened 2023-07-06
Fix EIP-712 improper handling of empty arrays

apaillier-ledger:
## Description

Fixes issues around empty arrays where the app would reject the APDUs and go back to the main screen mid-review.

Also fixes an issue with the NBGL code not properly cleaning up its EIP712 context in RAM.

## Changes include

- [x] Tests
- [ ] Documentation


src_features/signMessageEIP712/commands_712.c
```diff
@@ -38,6 +38,7 @@ void handle_eip712_return_code(bool success) {
 
     if (!success) {
         eip712_context_deinit();
+        ui_idle();
```
fbeutin-ledger, line 41: Not on stax ?

src_features/signMessageEIP712/path.c
```diff
@@ -143,14 +143,18 @@ static cx_sha3_t *get_last_hash_ctx(void) {
...
-static void finalize_hash_depth(uint8_t *hash) {
+static bool finalize_hash_depth(uint8_t *hash) {
     const cx_sha3_t *hash_ctx;
+    size_t hashed_bytes;
 
     hash_ctx = get_last_hash_ctx();
+    hashed_bytes = hash_ctx->blen;
     // finalize hash
     cx_hash((cx_hash_t *) hash_ctx, CX_LAST, NULL, 0, hash, KECCAK256_HASH_BYTESIZE);
     mem_dealloc(sizeof(*hash_ctx));  // remove hash context
```
fbeutin-ledger, line 156: Does this make sense if `hashed_bytes == 0` ?

src_features/signMessageEIP712/path.c
```diff
@@ -430,6 +442,8 @@ bool path_new_array_depth(const uint8_t *const data, uint8_t length) {
         return false;
     }
 
+    array_size = *data;
```
fbeutin-ledger, line 445: I really feel like eth app would benefit from a lot more comments, I have trouble understanding what huge functions like this are performing both, on their overhaul effect and in the actual implementation :/

src_features/signMessageEIP712/commands_712.c
apaillier-ledger: This code is common to all devices so it applies to Stax as well. :+1:

fbeutin-ledger (approved):

fbeutin-ledger (approved):

PR #488, opened 2023-11-06
improve plugin interface file

fbeutin-ledger:
## Description

improve plugin interface file

## Changes include

- [ ] Tests
- [X] Documentation

### DIFF
--- a/src/eth_plugin_interface.h
+++ b/src/eth_plugin_interface.h
@@ -1,3 +1,5 @@
+// clang-format off
+
 #ifndef _ETH_PLUGIN_INTERFACE_H_
 #define _ETH_PLUGIN_INTERFACE_H_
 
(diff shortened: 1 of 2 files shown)


src/eth_plugin_interface.h
```diff
@@ -7,30 +9,43 @@
 #include "tokens.h"
 #include "shared_context.h"
 
-// Interface version. To be updated everytime we introduce breaking changes to the plugin interface.
-typedef enum {
+/*************************************************************************************************
+ * Comments provided in this file are quick reminders on the usage of the plugin interface       *
+ * Reading the real plugin documentation is GREATLY recommended.                                 *
```
tdejoigny-ledger, line 14: we have also a good doc available on dev portal --> section : `plugin `and then `code-walkthrough`
where all the steps are described

src/eth_plugin_internal.h
```diff
@@ -21,6 +21,7 @@ bool U2BE_from_parameter(const uint8_t* parameter, uint16_t* value);
 bool U4BE_from_parameter(const uint8_t* parameter, uint32_t* value);
 
 typedef bool (*PluginAvailableCheck)(void);
+typedef void (*PluginCall)(int, void *);
```
fbeutin-ledger, line 24: @apaillier-ledger 
I've moved the function definition here, it doesn't seem to be used in any plugin code (and it's useless for them)

src/eth_plugin_interface.h
fbeutin-ledger: This is more generic, it has more its place in the README than in the comments on the interface file ?

apaillier-ledger (approved):

PR #522, opened 2024-01-24
fix: remove support for U2F

kewde:
## Description

This PR deprecates U2F as supported protocol.

> U2F is a technology that was hijacked by Ledger in 2016-2018 in order to achieve Web integrations of our devices. It was done at a time there was no other alternative and no way to communicate with "HID" or "WebUSB" technologies in a seemless manner.

U2F has been a thorn in the eye of WebUSB and has been hampering the user experience for the web.

[When `HAVE_IO_U2F` is enabled, it will also mangle the USB ProductID (PID)](https://github.com/LedgerHQ/ledger-secure-sdk/blob/5b3cda4e9667504fdf029f9cc95bd0897464d387/lib_stusb_impl/usbd_impl.c#L742-L760). If you switch into the Ethereum application, the USB PID changes from `0x5011` to `0x5015`  and the Chrome/Brave browser will identifiy it as a "different" device. This forces web developers to re-trigger the 'USB Pairing' screen. 

image
image

## Changes include

- [x] Breaking change (change that is not backwards-compatible and/or changes current functionality)
- [ ] Tests
- [ ] Documentation

## Breaking changes

Disable U2F.

## Additional comments

* BOLOS, Bitcoin and Solana apps both have a USB PID of `0x5011` and allow for smooth transitioning between them.


xchapron-ledger:
@apaillier-ledger I think we already discussed about it.
And If I remember correctly, at that time U2F was needed for Metamask usage. Do we know if this is still true?

> [When HAVE_IO_U2F is enabled, it will also mangle the USB ProductID (PID)](https://github.com/LedgerHQ/ledger-secure-sdk/blob/5b3cda4e9667504fdf029f9cc95bd0897464d387/lib_stusb_impl/usbd_impl.c#L742-L760). If you switch into the Ethereum application, the USB PID changes from 0x5011 to 0x5015 and the Chrome/Brave browser will identifiy it as a "different" device. This forces web developers to re-trigger the 'USB Pairing' screen.

Note that there is still a usb disconnexion / reconnexion when moving from an app to an another, even if there is no PID change (which might lead to a smooth handling from Chrome?). Though part of this behavior might change in future OS.

kewde:
I just checked to see for any previous context, and I've seen [one other attempt at removing this](https://github.com/LedgerHQ/app-ethereum/pull/344). 

* MetaMask browser extension uses [WebHID](https://github.com/MetaMask/metamask-extension/blob/b67f734c1fc55119e6c8a20649c80e5aaf824754/ui/components/app/ledger-instruction-field/ledger-instruction-field.js#L153-L171)
* MetaMask mobile does show [signs of  U2F](https://github.com/MetaMask/metamask-mobile/blob/2b66c53abbcf63440c179e8bebf64b0033c804d1/yarn.lock#L1976) but AFAIK it uses [bluetooth](https://github.com/MetaMask/metamask-mobile/pull/7757)

apaillier-ledger:
U2F was left activated in order to work with Firefox which does not support [WebHID](https://developer.mozilla.org/en-US/docs/Web/API/WebHID_API) (since it is not yet standard / still experimental).
However Firefox seems to have removed/changed something with the U2F support since version 114 and since then it has been broken and Metamask & Firefox teams have been pointing fingers at each other and nothing has been fixed.
Removing it from the app would mean that someone that has been happily using Metamask through an older version of Firefox would not be able to anymore. :thinking:

kewde:
I see, let's keep U2F then👍 

Do you happen to know [why the PID is being mutated based on which protocols it speaks](https://github.com/LedgerHQ/ledger-secure-sdk/blob/5b3cda4e9667504fdf029f9cc95bd0897464d387/lib_stusb_impl/usbd_impl.c#L742-L760)? Is that standardized somewhere, I haven't seen any other USB devices do this 😄 

Disabling U2F was just a means to an end, to get a stable PID with the other apps.

apaillier-ledger:
> Do you happen to know [why the PID is being mutated based on which protocols it speaks](https://github.com/LedgerHQ/ledger-secure-sdk/blob/5b3cda4e9667504fdf029f9cc95bd0897464d387/lib_stusb_impl/usbd_impl.c#L742-L760)? Is that standardized somewhere, I haven't seen any other USB devices do this 😄

I'm afraid I don't know the history behind this decision.

PR #534, opened 2024-02-08
EVM swap

apaillier-ledger:
## Description

Swap is now possible for EVM chains with the Ethereum app directly.
The app was missing one information lost from not going through the clone first : the chain ID.
Support was added for parsing the chain ID from the swap coin sub config (if present, so not a breaking change with current CAL).

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/eth_plugin_handler.c
+++ b/src/eth_plugin_handler.c
@@ -67,7 +67,9 @@ void eth_plugin_prepare_query_contract_UI(ethQueryContractUI_t *queryContractUI,
 
     queryContractUI->screenIndex = screenIndex;
     chain_id = get_tx_chain_id();
-    strlcpy(queryContractUI->network_ticker, get_displayable_ticker(&chain_id), MAX_TICKER_LEN);
+    strlcpy(queryContractUI->network_ticker,
+            get_displayable_ticker(&chain_id, chainConfig),
+            sizeof(queryContractUI->network_ticker));
     queryContractUI->title = title;
     queryContractUI->titleLength = titleLength;
     queryContractUI->msg = msg;
--- a/src/handle_get_printable_amount.c
+++ b/src/handle_get_printable_amount.c
@@ -7,32 +7,32 @@
 #include "common_utils.h"
 #include "uint256.h"
 #include "string.h"
+#include "network.h"
 
 void handle_get_printable_amount(get_printable_amount_parameters_t* params,
                                  chain_config_t* config) {
-    uint8_t decimals;
     char ticker[MAX_TICKER_LEN];
+    uint8_t decimals;
+    uint64_t chain_id = 0;
+
     memset(params->printable_amount, 0, sizeof(params->printable_amount));
     if (params->amount_length > 32) {
         PRINTF("Amount is too big, 32 bytes max but buffer has %u bytes", params->amount_length);
         return;
     }
 
-    // If the amount is a fee, its value is nominated in ETH even if we're doing an ERC20 swap
+    if (!parse_swap_config(params->coin_configuration,
+                           params->coin_configuration_length,
+                           ticker,
+                           &decimals,
+                           &chain_id)) {
+        PRINTF("Error while parsing config\n");
+        return;
+    }
+    // If the amount is a fee, the ticker should be the chain's native currency
     if (params->is_fee) {
-        uint8_t ticker_len = strnlen(config->coinName, sizeof(config->coinName));
-        memcpy(ticker, config->coinName, ticker_len);
-        ticker[ticker_len] = '\0';
+        strlcpy(ticker, get_displayable_ticker(&chain_id, config), sizeof(ticker));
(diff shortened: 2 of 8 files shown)


src/eth_plugin_handler.c
```diff
@@ -67,7 +67,9 @@ void eth_plugin_prepare_query_contract_UI(ethQueryContractUI_t *queryContractUI,
 
     queryContractUI->screenIndex = screenIndex;
     chain_id = get_tx_chain_id();
-    strlcpy(queryContractUI->network_ticker, get_displayable_ticker(&chain_id), MAX_TICKER_LEN);
+    strlcpy(queryContractUI->network_ticker,
```
fbeutin-ledger, line 70: Why not MAX_TICKER_LEN ?

src/handle_get_printable_amount.c
```diff
@@ -7,32 +7,32 @@
...
 
+    if (!parse_swap_config(params->coin_configuration,
+                           params->coin_configuration_length,
+                           ticker,
+                           &decimals,
+                           &chain_id)) {
+        PRINTF("Error while parsing config\n");
+        return;
+    }
     // If the amount is a fee, its value is nominated in ETH even if we're doing an ERC20 swap
```
fbeutin-ledger, line 32: update

src/handle_get_printable_amount.c
```diff
@@ -7,32 +7,32 @@
...
     if (params->is_fee) {
-        uint8_t ticker_len = strnlen(config->coinName, sizeof(config->coinName));
-        memcpy(ticker, config->coinName, ticker_len);
-        ticker[ticker_len] = '\0';
+        strlcpy(ticker, get_displayable_ticker(&chain_id, config), sizeof(ticker));
```
fbeutin-ledger, line 34: can get_displayable_ticker fail ?

src/swap_utils.c
```diff
@@ -38,6 +43,11 @@ bool parse_swap_config(const uint8_t *config, uint8_t config_len, char *ticker,
     if (config_len - offset < 1) {
         return false;
     }
-    *decimals = config[offset];
+    *decimals = config[offset++];
```
fbeutin-ledger, line 46: Do in 2 lines, this kind of increment inside operations is always a pain to debug / see errors

src/swap_utils.c
```diff
@@ -38,6 +43,11 @@ bool parse_swap_config(const uint8_t *config, uint8_t config_len, char *ticker,
...
+    // optional for retro-compatibility
+    if ((config_len - offset) >= sizeof(*chain_id)) {
```
fbeutin-ledger, line 48: Add logs

src/swap_utils.c
```diff
@@ -38,6 +43,11 @@ bool parse_swap_config(const uint8_t *config, uint8_t config_len, char *ticker,
     if (config_len - offset < 1) {
         return false;
     }
-    *decimals = config[offset];
+    *decimals = config[offset++];
+    // optional for retro-compatibility
```
fbeutin-ledger, line 47: Explain more in this comment what we receive in both cases

src/swap_utils.c
fbeutin-ledger, line 48: I think we can differentiate 3 cases:
- legacy size
- legacy size + sizeof(*chain_id)
- something else => error

src/swap_utils.h
```diff
@@ -19,4 +19,8 @@
 
 #include <stdint.h>
 
-bool parse_swap_config(const uint8_t* config, uint8_t config_len, char* ticker, uint8_t* decimals);
+bool parse_swap_config(const uint8_t* config,
+                       uint8_t config_len,
+                       char* ticker,
+                       uint8_t* decimals,
```
fbeutin-ledger, line 25: Not related, but I don't understand why sometimes the linter wants `int *var` and sometimes `int* var` ...

src/eth_plugin_handler.c
apaillier-ledger: So that if the size of `queryContractUI->network_ticker` ever changes, the strlcpy will adapt to it. :+1:

src/handle_get_printable_amount.c
apaillier-ledger: It cannot, there is always the backup from the given config.

src/swap_utils.c
apaillier-ledger: I would like it to ignore the leftover data if there is any (just like it was up until now), so we can keep on adding things at the end if the need ever arises.

src/swap_utils.h
apaillier-ledger: That is a good question.

src/handle_get_printable_amount.c
apaillier-ledger: A better solution will be looked at in a future PR :+1:

PR #563, opened 2024-04-03
Add yamllint CI and codespell CI

fbeutin-ledger:


.github/workflows/scripts_checking.yml
```diff
@@ -0,0 +1,30 @@
+name: Scripts quality CI
```
cedelavergne-ledger, line 1: The name does not reflect really the content, because it checks the source code also... Why not keeping the same name like in `app-boilerplate`: `Misspellings checks`?

Same for the yml script name: `misspellings_checks.yml`?

src_nbgl/ui_settings.c
```diff
@@ -34,19 +34,25 @@ static bool navCallback(uint8_t page, nbgl_pageContent_t* content) {
             switches[index++] =
                 (nbgl_layoutSwitch_t){.initState = N_storage.dataAllowed ? ON_STATE : OFF_STATE,
                                       .text = "Blind signing",
-                                      .subText = "Enable transaction blind\nsigning",
+                                      .subText =
+                                            "Enable transaction blind\n"
```
cedelavergne-ledger, line 38: Why do you need to split thetext like this?

.github/workflows/scripts_checking.yml
fbeutin-ledger: Done

src_nbgl/ui_settings.c
fbeutin-ledger: Updated the codespell version to avoid this workaround

cedelavergne-ledger (approved):

PR #565, opened 2024-04-08
B2CA-1558: Port Ethereum on Flex

cedelavergne-ledger:
Porting the App on Flex implies few intermediate steps. Here is what this PR includes:

- Adapt tree to **Flex**, fixing compilation errors
- Declare **Flex** in the manifest
- Adding **Flex** dedicated icons (App and Clones)
- Rename _glyphs_, removing `stax` prefix, because they are compatible with both **Stax** and **Flex**
- Port `nbgl_useCase_XXX` new API:
  - `nbgl_useCaseHomeAndSettings`
  - `nbgl_useCaseAddressReview`
  - `nbgl_useCaseReview`
  - `nbgl_useCaseGenericReview`
  - `nbgl_useCaseStreamingReview`
- Simplify the ragger test functions using `scenario_navigator.review_approve` and `scenario_navigator.review_reject`
- Update few _BAGL_ wording to avoid confusion during snapshot comparison (Change _Sign_ to _Review_ on 1st screen)
- Adapt the python client:
  - Adapt the `toggle_settings`, because Flex has less widgets on each page
  - Add command for new test `Perform Privacy Operation`
  - Remove useless delay (speculos is no more slow with Stax)
- Adapt the tests:
  - Check against **Flex** for `test_get_address.py`
  - Add tests for comand `Perform Privacy Operation` (can only run on device due to speculos limitation on curve `ed25519`)
  - Add test for comand `Confirm Parameter Selector`
- Add **Flex** snapshots

**Note**: There is still a remaining point, which is a limitation for now: _eip712_ is quiete complex, and needs more work to be ported to the new `nbgl_useCaseXXX` API. Thus, the corresponding ragger tests are skipped for **Flex**.

## Changes include

- [X] Tests
- [ ] Documentation
- [X] Other (for changes that might not fit in any category)


tdejoigny-ledger:
As part of Flex porting it could be good also to update this [file](https://github.com/LedgerHQ/app-ethereum/blob/develop/src_plugin_sdk/standard_plugin.mk) in order to prepare the plugins flex porting

apaillier-ledger (approved):

PR #608, opened 2024-07-24
EIP-712 filters counter fix + reactivation

apaillier-ledger:
## Description

Re-activate & fix the EIP-712 filters counter which was disabled in #582, now properly counts unique filters.
The proper handling of filtered path within empty arrays will come later since it also requires a ledgerjs modification.

Uses yet again a bit more RAM (200 bytes) so increased the RAM buffer to make up for it.
Now on Nanos, does not let the user go back to the review page once he has reached the approve/reject.

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src_features/signMessageEIP712/ui_logic.c
+++ b/src_features/signMessageEIP712/ui_logic.c
@@ -18,6 +18,7 @@
 #include "commands_712.h"
 #include "common_ui.h"
 #include "uint_common.h"
+#include "filtering.h"
 
 #define AMOUNT_JOIN_FLAG_TOKEN (1 << 0)
 #define AMOUNT_JOIN_FLAG_VALUE (1 << 1)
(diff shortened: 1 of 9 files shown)


apaillier-ledger:
CI is red on Stax & Flex because of this Speculos issue : https://github.com/LedgerHQ/speculos/issues/508
The computation has always a result of 0x00000000, so for each subsequent EIP-712 filter it finds an already matching one in RAM, therefore only counting the first one and complaining at the end that there are missing filters.
Works on physical devices. :+1:

src_features/signMessageEIP712/ui_logic.c
```diff
@@ -739,20 +738,17 @@ void ui_712_queue_struct_to_review(void) {
...
-            }
-        }
+ * @return if the counter could be incremented
+ */
+bool ui_712_filters_counter_incr(void) {
+    if (ui_ctx->filters_received > ui_ctx->filters_to_process) {
+        return false;
     }
+    ui_ctx->filters_received += 1;
+    ;
```
tdejoigny-ledger, line 750: extra ;

tdejoigny-ledger (approved):

src_features/signMessageEIP712/ui_logic.c
apaillier-ledger: Good catch

tdejoigny-ledger (approved):

PR #626, opened 2024-08-12
Add and update PulseChain Networks

bretep:
## Description

Please provide a detailed description of what was done in this PR.
(And mentioned if linked to an issue [docs](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue))

## Changes include

- [ ] Tests
- [ ] Documentation

## Additional comments

Support for PulseChain was missed during updates earlier this year. I've added it back following the new patterns.

Ledger Live integration: https://github.com/LedgerHQ/ledger-live/pull/7607

### DIFF
--- a/glyphs/chain_943_64px.gif
+++ b/glyphs/chain_943_64px.gif
@@ -1,0 +1,1 @@
+chain_369_64px.gif
--- a/src/network.c
+++ b/src/network.c
@@ -52,6 +52,7 @@ static const network_info_t NETWORK_MAPPING[] = {
     {.chain_id = 369, .name = "PulseChain", .ticker = "PLS"},
     {.chain_id = 592, .name = "Astar", .ticker = "ASTR"},
     {.chain_id = 888, .name = "Wanchain", .ticker = "WAN"},
+    {.chain_id = 943, .name = "PulseChain Testnet", .ticker = "tPLS"},
     {.chain_id = 1030, .name = "Conflux", .ticker = "CFX"},
     {.chain_id = 1088, .name = "Metis Andromeda", .ticker = "METIS"},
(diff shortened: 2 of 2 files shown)


apaillier-ledger:
Hello @bretep, thank you for this PR.

We now don't allow clone applications (with separate .mk files) that only use Ethereum's derivation path, since they will work just fine directly with the Ethereum app directly.

If you could update your PR by :
* removing the two .mk files you added
* removing the .gif files you added in icons/
* since the glyphs for the mainnet and testnet are the same, to save space in both the repository and the app itself, create a symbolic link instead so that `chain_943_64px.gif` points to `chain_369_64px.gif`

Thank you! :pray:

bretep:
@apaillier-ledger, Thank you for that excellent review and feedback.

I've done the following:

- Removed the two `.mk` files
- Removed the `.gif` icons located in `./icons`
- Symlinked the chain's `.gif` in `./glyphs` to save space
- Amended the previous commit to keep this a single commit
- Force pushed

Please let me know if I need to fix anything else. I appreciate your help.

bretep:
@apaillier-ledger I guess the only concern I have is `DEFINES += HAVE_ETH2`. Because PulseChain has a beacon chain and uses the same BLS curve as Ethereum, will validators still be able to use the Ethereum App to make validator addresses?

Most other Ethereum chains do not have a beacon chain; they use ETH1. 

Reference:
https://beacon.pulsechain.com
https://launchpad.pulsechain.com

apaillier-ledger:
@bretep, it should behave with the Ethereum app the same way as it would have with the separate clone app.
Let us know if you have issues between your beacon chain and the Ethereum app, since the ETH2 support was initially developped for the mainnet one.

apaillier-ledger (approved):

PR #632, opened 2024-08-27
Fix EIP-712 filtering on empty array

apaillier-ledger:
## Description

For filters targeting fields within an array that turned out to be empty, the app would never receive these fields implementation (because they don't exist), same for their corresponding filters, so the app would at the end of the review flow complain about missing filters.

## Changes include

- [x] Tests
- [x] Documentation


src_features/signMessageEIP712/filtering.c
```diff
@@ -28,33 +28,41 @@
  * Reconstruct the field path and hash it
  *
  * @param[in] hash_ctx the hashing context
+ * @param[in] discarded if the filter targets a field that is does not exist (within an empty array)
```
cedelavergne-ledger, line 31: minor: `it` instead of `is`

cedelavergne-ledger (approved):

PR #635, opened 2024-08-29
B2CA-1566: Dynamic network management

cedelavergne-ledger:
## Description

Add feature (new APDUs) to inject a network config at runtime.
This configuration is not stored on the device, just used in RAM for the running session.
Up to 2 configurations can be used.
Signature on the received payload is mandatory, using a new dedicated key (LedgerPKI supported).

The doc is updated accordingly, which also fixes https://ledgerhq.atlassian.net/browse/B2CA-612

_**Note**_: Still to remove the Boilerplate configuration in `client/src/ledger_app_clients/ethereum/client.py`, line 519-529, just here to showcase the behavior.

## Changes include

- [ ] Tests
- [X] Documentation


.github/workflows/build_and_functional_tests.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     uses: LedgerHQ/ledger-app-workflows/.github/workflows/reusable_build.yml@v1
     with:
       upload_app_binaries_artifact: "ragger_elfs"
-      flags: "CAL_TEST_KEY=1 TRUSTED_NAME_TEST_KEY=1 SET_PLUGIN_TEST_KEY=1 NFT_TEST_KEY=1"
+      flags: "CAL_TEST_KEY=1 TRUSTED_NAME_TEST_KEY=1 SET_PLUGIN_TEST_KEY=1 NFT_TEST_KEY=1 DYNAMIC_NETWORK_TEST_KEY=1"
```
apaillier-ledger, line 35: If it uses the CAL key, we could just re-use CAL_TEST_KEY

client/src/ledger_app_clients/ethereum/keychain.py
```diff
@@ -12,6 +12,7 @@ class Key(Enum):
     TRUSTED_NAME = auto()
     SET_PLUGIN = auto()
     NFT = auto()
+    DYNAMIC_NETWORK = auto()
```
apaillier-ledger, line 15: Not needed if it uses the CAL key.

src/network_dynamic.c
apaillier-ledger, line 1: To be aligned with other features in the app that are exposed over APDUs, I would move this to `src_features/`

doc/ethapp.adoc
```diff
@@ -1089,6 +1070,91 @@ _Output data_
...
+>
+>  - Its value is the hex string generated by the script `<SDK_PATH>/lib_nbgl/tools/icon2glyph.py`, with parameter `--hexbitmap`
+
+The signature is mandatory, and is computed on the full payload data, using `CX_CURVE_SECP256K1`.
+
+#### Coding
+
+_Command_
+
+[width="80%"]
```
apaillier-ledger, line 1115: To be aligned with other APDUs, P1 & P2 should be switched (with first/following chunks being in P1).

doc/ethapp.adoc
```diff
@@ -1089,6 +1070,91 @@ _Output data_
...
+|==============================================================
+| *CLA*     | *INS*    | *P1*                       | *P2*                     | *LC*     | *Le*
+.3+| E0  .3+| 30       | 00 : Network configuration | <slot>                   | variable | variable
```
apaillier-ledger, line 1118: The fact that the wallet has to explicitly handle which slot it wants to use seems way overkill.
It could simply be a ring-buffer like it is for ERC-20 tokens.

doc/ethapp.adoc
```diff
@@ -1089,6 +1070,91 @@ _Output data_
...
+|==============================================================
+
+_Input data_
+
+##### If P1 == Network configuration
+
+[width="80%"]
+|==========================================
+| *Description*         | *Length (byte)*
+| TLV payload data      | variable
```
apaillier-ledger, line 1133: Are we sure this command will never have to be split across multiple APDUs ?

doc/ethapp.adoc
```diff
@@ -1089,6 +1070,91 @@ _Output data_
...
+|==========================================
+
+##### If P1 == Network Icon
+
+[width="80%"]
+|==========================================
+| *Description*         | *Length (byte)*
+| Icon data             | variable
```
apaillier-ledger, line 1141: This should be prefixed by its size, to handle multi-APDUs payloads.

src/network_dynamic.c
```diff
@@ -0,0 +1,654 @@
...
+typedef struct {
+    uint16_t received_size;
+    uint16_t expected_size;
+    uint16_t slot;
+} network_payload_t;
+
+// Global variable to store the network icons
+typedef struct {
+    uint8_t bitmap[MAX_ICON_LEN];
+    uint8_t hash[CX_SHA256_SIZE];
```
apaillier-ledger, line 61: Since we already received the hash along with the network configuration, couldn't we do something like this to match the two ? `icon_hash_from_network_config == sha256(icon_bitmap)`
It would save on RAM at the (small) cost of a little runtime computation.

src/public_keys.h
```diff
@@ -105,6 +105,31 @@ static const uint8_t LEDGER_NFT_SELECTOR_PUBLIC_KEY[] = {
 #endif
 };
 
+static const uint8_t DYNAMIC_NETWORK_PUB_KEY[] = {
```
apaillier-ledger, line 108: If this uses the CAL key, might as well reuse it instead of duplicating it.

apaillier-ledger (changes requested):

.github/workflows/build_and_functional_tests.yml
cedelavergne-ledger: Ok. Fixed

client/src/ledger_app_clients/ethereum/keychain.py
cedelavergne-ledger: Removed

src/public_keys.h
cedelavergne-ledger: Removed

doc/ethapp.adoc
cedelavergne-ledger: Yes, this is the reason to use a hash and not the icon directly. Splitting over several chunk would have impacts on RAM footprint also...

doc/ethapp.adoc
cedelavergne-ledger: In fact, this is not needed, because the image bitmap bytes array starts by a header which indicate the full size. So it is already retrieved and checked here, on the 1st chunk.

src/network_dynamic.c
cedelavergne-ledger: Done

src/network_dynamic.c
cedelavergne-ledger: This is how it works :wink:

doc/ethapp.adoc
cedelavergne-ledger: Fixed

doc/ethapp.adoc
cedelavergne-ledger: Updated

src_features/networkDynamic/network_dynamic.c
```diff
@@ -0,0 +1,644 @@
...
+static uint16_t parse_tlv(const uint8_t *data, uint8_t length) {
+    uint16_t offset = 0;
+    uint16_t field_tag = 0;
+    uint16_t field_len = 0;
+    uint16_t sw = APDU_RESPONSE_INTERNAL_ERROR;
+    uint16_t tag_start_off;
+    s_sig_ctx sig_ctx = {0};
+
+    // Set the current slot
+    g_current_slot = (g_current_slot + 1) % MAX_DYNAMIC_NETWORKS;
```
apaillier-ledger, line 484: This should be put at the end of the function, in the same `if` as the call to `print_network_info` so that we can ensure the first index to be used is 0, then 1, and so on. :+1:

apaillier-ledger (changes requested):

src_features/networkDynamic/network_dynamic.c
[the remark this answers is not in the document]
cedelavergne-ledger: :+1: 
Done

src_features/networkDynamic/network_dynamic.c
```diff
@@ -0,0 +1,644 @@
...
+
+// Signature context structure
+typedef struct {
+    uint8_t sig_size;
+    const uint8_t *sig;
+    cx_sha256_t hash_ctx;
+} s_sig_ctx;
+
+// Global variable to store the current slot
+uint8_t g_current_slot;
```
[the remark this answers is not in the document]
cedelavergne-ledger: ok, done

src_features/networkDynamic/network_dynamic.c
cedelavergne-ledger: Need to keep it here, at the begining of the function, to ensure the icon, sent separately will use the same slot.
We could use another slot_index, but the code would be less readable.

apaillier-ledger (approved):

PR #639, opened 2024-09-06
EIP-712 amount-joins with missing token information

apaillier-ledger:
## Description

Now instead of considering this an error, shows the amount-join with the little information it has like it does for transactions (with `???` for ticker).

## Changes include

- [x] Tests
- [ ] Documentation


client/src/ledger_app_clients/ethereum/eip712/InputData.py
```diff
@@ -17,6 +17,7 @@
 # global variables
 app_client: EthAppClient = None
 filtering_paths: dict = {}
+filtering_tokens: list[dict] = list()
```
cedelavergne-ledger, line 20: Minor, but initializing with simply `[]` instead of `list()` seems more python oriented.

cedelavergne-ledger (approved):

cedelavergne-ledger (dismissed):

PR #651, opened 2024-09-20
Trusted names v2 + EIP-712 trusted name filtering

apaillier-ledger:
## Description

* Added support for V2 trusted name TLV payloads (still backwards compatible)
* Added a new EIP-712 filter for trusted names

## Changes include

- [x] Tests
- [x] Documentation


src/main.c
```diff
@@ -320,8 +320,8 @@ void app_main(void) {
 
 static void init_coin_config(chain_config_t *coin_config) {
     memset(coin_config, 0, sizeof(chain_config_t));
-    strcpy(coin_config->coinName, CHAINID_COINNAME);
-    coin_config->chainId = CHAIN_ID;
+    strcpy(coin_config->coinName, APP_TICKER);
```
cedelavergne-ledger, line 323: Should we also change the field `coinName` by something like `coinTicker`?

apaillier-ledger: We could, but in a later PR. :+1:

PR #697, opened 2024-12-13
add Odyssey Chain

IntegralTeam:
Added support of Odyssey Chain according to this documentation - https://developers.ledger.com/docs/device-app/integration/how-to/clone#a---using-the-4460-path


vforgeoux-ledger:
Taken into account in this PR : https://github.com/LedgerHQ/app-ethereum/pull/713

PR #699, opened 2024-12-16
Generic clear-signing

apaillier-ledger:
## Description

Added new feature for clear-signing that will replace existing plugins (for the most part).

## Changes include

- [x] Tests
- [x] Documentation


src_features/generic_tx_parser/cmd_field.c
```diff
@@ -0,0 +1,50 @@
...
+
+static bool handle_tlv_payload(const uint8_t *payload, uint16_t size, bool to_free) {
+    s_field field;
+    s_field_ctx ctx;
+
+    explicit_bzero(&field, sizeof(field));
+    explicit_bzero(&ctx, sizeof(ctx));
+    ctx.field = &field;
+    if (!tlv_parse(payload, size, (f_tlv_data_handler) &handle_field_struct, &ctx)) return false;
+    if (cx_hash_no_throw(get_fields_hash_ctx(), 0, payload, size, NULL, 0) != CX_OK) return false;
```
cedelavergne-ledger, line 21: We don't care here is the payload contains a Signature tag?

src_features/generic_tx_parser/cmd_tx_info.c
```diff
@@ -0,0 +1,128 @@
...
+
+    explicit_bzero(&ctx, sizeof(ctx));
+    ctx.tx_info = g_tx_info;
+    cx_sha256_init((cx_sha256_t *) &ctx.struct_hash);
+    if (!tlv_parse(payload, size, (f_tlv_data_handler) &handle_tx_info_struct, &ctx)) return false;
+    if (to_free) mem_dealloc(sizeof(size));
+    if (!verify_struct(&ctx)) {
+        return false;
+    }
+    if (cx_sha3_init_no_throw(&hash_ctx, 256) != CX_OK) {
```
cedelavergne-ledger, line 83: both sha3 and sha256 hash ?

src_features/generic_tx_parser/cmd_tx_info.c
```diff
@@ -0,0 +1,128 @@
...
+    return true;
+}
+
+static bool handle_tlv_payload(const uint8_t *payload, uint16_t size, bool to_free) {
+    s_tx_info_ctx ctx;
+
+    explicit_bzero(&ctx, sizeof(ctx));
+    ctx.tx_info = g_tx_info;
+    cx_sha256_init((cx_sha256_t *) &ctx.struct_hash);
+    if (!tlv_parse(payload, size, (f_tlv_data_handler) &handle_tx_info_struct, &ctx)) return false;
```
cedelavergne-ledger, line 78: Where do you handle the signature tag?

cedelavergne-ledger (changes requested):
Seems that there is some duplicates in the different handlers (like handle_version, handle_id, maybe handle_struct...)
I think the Info icon could be used from the SDK ?

Sorry, but I didn't go deeper in each file, this PR is quite complex :stuck_out_tongue_winking_eye:

client/src/ledger_app_clients/ethereum/command_builder.py
```diff
@@ -427,3 +429,23 @@ def provide_network_information(self,
                 icon = icon[0xff:]
                 p1 = P1Type.FOLLOWING_CHUNK
         return chunks
+
+    def common_tlv_serialize(self, tlv_payload: bytes, ins: InsType) -> list[bytes]:
```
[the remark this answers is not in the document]
apaillier-ledger: Yes, the name might not be optimal, but it just takes a raw TLV payload and prefixes it with its length. (also sets the P1 depending if it is the first chunk or not. :+1:

src_features/generic_tx_parser/cmd_field.c
apaillier-ledger: This payload is not signed, since the signed TX info payload contains a hash of all the field structs. :+1:

src_features/generic_tx_parser/cmd_tx_info.c
apaillier-ledger: In the function `handle_signature` of file `gtp_tx_info.c`

src_features/generic_tx_parser/cmd_tx_info.c
apaillier-ledger: Yes, sha(2)-256 is used to verify the tx info struct signature, and the sha3-256 is initialized to verify the hash of all the field structs.

src_features/generic_tx_parser/gtp_field_table.c
```diff
@@ -0,0 +1,99 @@
...
+    if (entry != NULL) {
+        entry->key = (char *) ptr;
+    }
+    ptr += key_len;
+    if (entry != NULL) {
+        entry->value = (char *) ptr;
+    }
+    ptr += value_len;
+    return ptr;
+}
```
bboilot-ledger, line 42: Not sure to understand the behavior when `entry == NULL` there. Do you want to be able to get a pointer to the next entry without copying the current value? However, if `entry` is indeed NULL, the line `ptr += sizeof(entry->type);` will still be executed resulting in a segmentation fault. Since we're leveraging the memory allocator, could we just store the next object pointer in the previous struct like a linked list? Seems simpler and better in term of performance.

src_features/generic_tx_parser/gtp_field_table.c
```diff
@@ -0,0 +1,99 @@
...
+
+void field_table_cleanup(void) {
```
bboilot-ledger, line 44: Can you confirm that no other allocations are made between the calls to  `add_to_field_table` and the call to `field_table_cleanup`?

src_features/generic_tx_parser/gtp_field_table.c
```diff
@@ -0,0 +1,99 @@
...
+    size_t size;
+} s_field_table;
+
+static s_field_table g_table;
+
+void field_table_init(void) {
+    explicit_bzero(&g_table, sizeof(g_table));
+}
+
+const uint8_t *get_next_table_entry(const uint8_t *ptr, s_field_table_entry *entry) {
```
bboilot-ledger, line 21: We could check `ptr` value to ensure it's not NULL.

src_features/generic_tx_parser/gtp_field_table.c
```diff
@@ -0,0 +1,99 @@
...
+    ptr += key_len;
+    if (entry != NULL) {
+        entry->value = (char *) ptr;
+    }
+    ptr += value_len;
+    return ptr;
+}
+
+void field_table_cleanup(void) {
+    const uint8_t *ptr = g_table.start;
```
bboilot-ledger, line 45: We could check ptr value to ensure it's not NULL.

src_features/generic_tx_parser/gtp_field_table.c
```diff
@@ -0,0 +1,99 @@
...
+
+    for (size_t i = 0; i < g_table.size; ++i) {
+        ptr = get_next_table_entry(ptr, NULL);
+    }
+    mem_dealloc(ptr - g_table.start);
+}
+
+bool add_to_field_table(e_param_type type, const char *key, const char *value) {
```
bboilot-ledger, line 53: We could check that both `key` and `value` are not null. Also can you confirm that there is no limitation on there size (meaning that the limitation is fixed by the allocator I guess)?

src_features/generic_tx_parser/gtp_field_table.c
bboilot-ledger, line 44: Shouldn't cleanup also reset `g_table` to avoid use after free?

bboilot-ledger (changes requested):
Here are few comments mostly on the code not covered by the fuzzer.

src_features/generic_tx_parser/gtp_field_table.c
apaillier-ledger: Indeed, no other allocation. :+1:

apaillier-ledger: It could, but it would be redundant with the `field_table_init` call from `cmd_tx_info.c`, and we can not get into these functions without having received a valid TX info beforehand.

src_features/generic_tx_parser/gtp_field_table.c
bboilot-ledger: Oh! This is true, my bad!

src_features/generic_tx_parser/gtp_field_table.c
bboilot-ledger: My point is that with the current flow, when we cleanup, we only reuse the `g_table` struct after a `field_table_init` so it's fine. But if somehow `g_table` is reused after `field_table_cleanup`, the pointer to the data will still be valid and usable. However, this data has been freed from the allocator and could be altered. 
If you decide to keep that behavior, could you add a comment saying that if `field_table_cleanup` is called, `field_table_init` must be called before using any field_table related APIs again.

src_features/generic_tx_parser/gtp_field_table.c
apaillier-ledger: Indeed no limitation.

cedelavergne-ledger (approved):

bboilot-ledger (approved):
:tada:

PR #700, opened 2024-12-16
Add 1st fuzzing

cedelavergne-ledger:
Based on Donjon review https://ledgerhq.atlassian.net/browse/DSA-953


.clusterfuzzlite/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+# Base image with clang toolchain
+FROM gcr.io/oss-fuzz-base/base-builder:v1
+
+# Copy the project's source code.
+COPY . $SRC/app-ethereum
+COPY --from=LITE_BUILDER /opt/ledger-secure-sdk $SRC/app-ethereum/BOLOS_SDK
+
+# Add the ethereum-plugin-sdk submodule
+RUN git clone https://github.com/LedgerHQ/ethereum-plugin-sdk.git $SRC/app-ethereum/ethereum-plugin-sdk
```
bboilot-ledger, line 11: It should be the job of clusterfuzzlite action to recursively clone the repository. However, it doesn't support it currently: https://github.com/google/clusterfuzzlite/issues/107

bboilot-ledger:
Hello :wave: 
I have added the other fuzzers done during review and merged everything in one harness. Also cleaned up few things in the cmake.
Added a script to run the fuzzer locally and compute the global coverage.
A few bugs reported by the fuzzer have been fixed (I am not using the memory fuzzer much but clusterfuzzlite runs it everytime).
There is still at least one issue on the memory fuzzer fro the trusted names feature that needs to be investigated.

tests/fuzzing/src/mock.c
```diff
@@ -0,0 +1,171 @@
+#include <string.h>
+#include <stdlib.h>
+
+#include "cx_errors.h"
+#include "cx_sha256.h"
+#include "cx_sha3.h"
+
+#define explicit_bzero(X, Y) memset_s(X, 0, Y)
```
cedelavergne-ledger, line 8: Why defining this macro, rather than calling directly `memset_s` functions in the different other functions in this file?

bboilot-ledger: Indeed, I initially thought that I would have to globally replace explicit_bzero implementation for the fuzzing target. But  fixing the mock was enough. So, I removed the macro following your recommendation.

apaillier-ledger (approved):

PR #709, opened 2025-01-08
Unify TLV parsing

apaillier-ledger:
## Description

Make all the existing features using TLV use the new parser that came with the generic parser.

## Changes include

- [ ] Tests
- [ ] Documentation


cedelavergne-ledger:
During the tests, I noticed errors with Ledger-PKI which has not been updated after being changed from `Domain Name` to `Trusted Name`. The following APDU certificate metadata should have been updated: `Trusted Name` and `Key Usage`.

PR #712, opened 2025-01-20
calldata compression

apaillier-ledger:
## Description

Simple compression of the stored calldata in RAM during generic clear-signing.

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/mem.c
+++ b/src/mem.c
@@ -15,13 +15,15 @@
 #define SIZE_MEM_BUFFER 10240
 
 static uint8_t mem_buffer[SIZE_MEM_BUFFER];
-static size_t mem_idx;
+static uint16_t mem_idx;
+static uint16_t mem_rev_idx;
 
 /**
  * Initializes the memory buffer index
  */
 void mem_init(void) {
     mem_idx = 0;
+    mem_rev_idx = 0;
 }
 
 /**
@@ -42,7 +44,7 @@ void mem_reset(void) {
  */
 void *mem_alloc(size_t size) {
     // Buffer exceeded
-    if ((mem_idx + size) > sizeof(mem_buffer)) {
+    if ((mem_idx + size) > (sizeof(mem_buffer) - mem_rev_idx)) {
         PRINTF("Error: mem_alloc(%u) failed!\n", size);
         return NULL;
     }
@@ -62,6 +64,37 @@ void mem_dealloc(size_t size) {
         mem_idx = 0;
     } else {
         mem_idx -= size;
+    }
+}
+
+/**
+ * Same as \ref mem_alloc but in reverse
+ *
+ * @param[in] size Requested allocation size in bytes
+ * @return Allocated memory pointer; \ref NULL if not enough space left.
+ */
+void *mem_rev_alloc(size_t size) {
+    // Buffer exceeded
+    if ((sizeof(mem_buffer) - (mem_rev_idx + size)) < mem_idx) {
+        PRINTF("Error: mem_rev_alloc(%u) failed!\n", size);
+        return NULL;
+    }
+    mem_rev_idx += size;
+    return &mem_buffer[sizeof(mem_buffer) - mem_rev_idx];
+}
+
+/**
+ * Same as \ref mem_dealloc but in reverse
+ *
+ * @param[in] size Requested deallocation size in bytes
+ */
+void mem_rev_dealloc(size_t size) {
+    // More than is already allocated
+    if (size > mem_rev_idx) {
+        PRINTF("Warning: mem_rev_dealloc(%u) with a value larger than allocated!\n", size);
+        mem_rev_idx = 0;
+    } else {
+        mem_rev_idx -= size;
     }
 }
 
--- a/src/mem_utils.c
+++ b/src/mem_utils.c
@@ -37,6 +37,22 @@ char *mem_alloc_and_format_uint(uint32_t value, uint8_t *const length) {
 }
 
 /**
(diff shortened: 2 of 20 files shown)


src/mem_utils.c
```diff
@@ -46,14 +62,7 @@ char *mem_alloc_and_format_uint(uint32_t value, uint8_t *const length) {
...
 void *mem_alloc_and_align(size_t size, size_t alignment) {
-    uint8_t align_diff = (uintptr_t) mem_alloc(0) % alignment;
-
-    if (align_diff > 0)  // alignment needed
-    {
-        if (mem_alloc(alignment - align_diff) == NULL) {
-            return NULL;
-        }
-    }
+    mem_align(alignment);
```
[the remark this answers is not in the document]
apaillier-ledger: Indeed, while not great, it is only used in the EIP-712 feature, and that features calls `mem_reset` when it finishes.
It will be fixed altogether when we move to an actual memory allocator instead of that simple stack.

src_features/generic_tx_parser/gtp_data_path.c
```diff
@@ -308,4 +324,10 @@ bool data_path_get(const s_data_path *data_path, s_parsed_value_collection *coll
     return true;
 }
 
+void data_path_cleanup(const s_parsed_value_collection *collection) {
+    for (int i = 0; i < collection->size; ++i) {
+        mem_rev_dealloc(collection->value[i].size);
```
cedelavergne-ledger, line 329: I don't undetstand why you dealloc using `size` field, whereas the alloc was done using `length` field?

src_features/generic_tx_parser/gtp_param_nft.c
```diff
@@ -53,45 +53,53 @@ bool handle_param_nft_struct(const s_tlv_data *data, s_param_nft_context *contex
...
-        if ((asset = (const extraInfo_t *) get_asset_info_by_addr(addr_buf)) == NULL) {
-            return false;
-        }
-        if (!uint256_to_decimal(ids.value[i].ptr, ids.value[i].length, tmp, sizeof(tmp))) {
-            return false;
-        }
-        snprintf(buf, buf_size, "%s #%s", asset->nft.collectionName, tmp);
-        if (!add_to_field_table(PARAM_TYPE_NFT, name, buf)) {
-            return false;
+    if ((ret = value_get(&param->collection, &collections))) {
```
cedelavergne-ledger, line 66: The code is correct, but for readability, using `goto` would be easier. Just a question of personal choice :wink:

src_features/generic_tx_parser/gtp_param_token_amount.c
```diff
@@ -113,73 +113,80 @@ static bool match_native(const uint8_t *addr, const s_param_token_amount *param)
...
+    const char *ticker = g_unknown_ticker;
+    uint8_t decimals = 0;
 
-    if (!value_get(&param->value, &collec_value)) {
-        return false;
-    }
     if (param->has_token) {
-        if (!value_get(&param->token, &collec_token)) {
-            return false;
+        buf_shrink_expand(token->ptr, value->length, addr_buf, sizeof(addr_buf));
```
cedelavergne-ledger, line 131: The 2nd parameter was token->length??

src_features/generic_tx_parser/gtp_data_path.c
apaillier-ledger: Initially `size == length`, if a slice is applied to the value, then `length` becomes smaller but `size` remains the same.
So when we `dealloc`, we need to use `size`.

src/mem.c
```diff
@@ -65,4 +67,35 @@ void mem_dealloc(size_t size) {
...
+/**
+ * Same as \ref mem_dealloc but in reverse
+ *
+ * @param[in] size Requested deallocation size in bytes
+ */
+void mem_rev_dealloc(size_t size) {
+    // More than is already allocated
+    if (size > mem_rev_idx) {
+        PRINTF("Warning: mem_rev_dealloc(%u) with a value larger than allocated!\n", size);
+        mem_rev_idx = 0;
```
cedelavergne-ledger, line 95: Here, setting to 0, we directly deallocate the whole reverse buffer, ans thus, we'll probably have side effects like data corruption,.
Would it be better to not deallocate (and accpet a memory leak)?

src_features/generic_tx_parser/gtp_param_token_amount.c
apaillier-ledger: Very good catch! :exploding_head:

src_features/generic_tx_parser/gtp_param_token_amount.c
```diff
@@ -113,73 +113,80 @@ static bool match_native(const uint8_t *addr, const s_param_token_amount *param)
...
-    for (int i = 0; i < collec_value.size; ++i) {
+    if (!add_to_field_table(PARAM_TYPE_AMOUNT, name, buf)) {
+        return false;
+    }
+    return true;
+}
+
+bool format_param_token_amount(const s_param_token_amount *param, const char *name) {
+    bool ret;
+    s_parsed_value_collection collec_value;
```
cedelavergne-ledger, line 164: should be init with `{0}`

apaillier-ledger: Indeed, I've added it to all parsed value collections. :+1:

cedelavergne-ledger (approved):

PR #724, opened 2025-01-28
B2CA-1881:  Add web3check feature

cedelavergne-ledger:
## Description

- Add setting to enable web3check

## Changes include

- [X] Tests
- [X] Documentation


src_features/provideTxSimu/ui_getTxSimulation.c
apaillier-ledger, line 1: NBGL code should go in `src_nbgl/`

src_features/provideTxSimu/cmd_getTxSimulation.c
apaillier-ledger, line 1: Directory & file names should use snake_case and not CamelCase to be more in line with new code.

client/src/ledger_app_clients/ethereum/keychain.py
```diff
@@ -14,6 +14,7 @@ class Key(Enum):
     NFT = auto()
     CALLDATA = auto()
     NETWORK = auto()
+    W3C = auto()
```
apaillier-ledger, line 17: W3C stands for something else and is misleading, should be renamed everywhere to something else : 
* web3_checks / web3checks
* WEB3_CHECKS / WEB3CHECKS

client/src/ledger_app_clients/ethereum/client.py
```diff
@@ -606,3 +632,81 @@ def provide_transaction_info(self, payload: bytes) -> RAPDU:
         for chunk in chunks[:-1]:
             self._exchange(chunk)
         return self._exchange(chunks[-1])
+
+    def opt_in_tx_simulation(self):
+        return self._exchange_async(self._cmd_builder.opt_in_tx_simulation())
+
+    def _prepare_tx_simulation(self,
```
apaillier-ledger, line 639: I tend to prefer defining a Python class and in the tests building the object then calling the client function by passing the result of its `.serialize()` method, like how it is done for GCS tests.
This way if the TLV payload evolves, we won't have to change the client functions, but only the class itself.

apaillier-ledger (changes requested):

src_features/provideTxSimu/ui_getTxSimulation.c
cedelavergne-ledger: Ok, file moved

client/src/ledger_app_clients/ethereum/keychain.py
cedelavergne-ledger: Done, renamed

client/src/ledger_app_clients/ethereum/client.py
cedelavergne-ledger: Done. Added a client for tx_simu, and re-adapt the different tests

apaillier-ledger (approved):

PR #738, opened 2025-02-21
Use new setup feature from ragger

fbeutin-ledger:


tests/ragger/conftest.py
```diff
@@ -18,33 +18,14 @@
 # ragger.configuration.OPTIONAL_CONFIGURATION
 # Please refer to ragger/conftest/configuration.py for their descriptions and accepted values
 
+configuration.OPTIONAL.ALLOWED_SETUPS = ["default", "lib_mode"]
```
cedelavergne-ledger, line 21: It seems `default` will be added by `base_conftest.py` if not set here, no?

cedelavergne-ledger (approved):

tests/ragger/conftest.py
fbeutin-ledger: Yes, but I think it's more explicit to have the two written

PR #748, opened 2025-03-10
GCS PARAM_TRUSTED_NAME sender address support

apaillier-ledger:
## Description

Add support for a new optional field in the GCS formatter for trusted names.

## Changes include

- [x] Tests
- [ ] Documentation


src_features/generic_tx_parser/gtp_param_trusted_name.c
```diff
@@ -7,12 +7,15 @@
...
 #include "shared_context.h"
+#include "feature_signTx.h"  // get_public_key
+#include "apdu_constants.h"
 
 enum {
     TAG_VERSION = 0x00,
     TAG_VALUE = 0x01,
     TAG_TYPES = 0x02,
     TAG_SOURCES = 0x03,
+    TAG_SENDER_ADDR = 0x04,
```
cedelavergne-ledger, line 18: Same comment as python file above :wink:

client/src/ledger_app_clients/ethereum/gcs.py
```diff
@@ -428,6 +435,9 @@ def serialize(self) -> bytes:
         for source in self.sources:
             sources.append(source)
         payload += format_tlv(0x03, sources)
+        if self.sender_addrs is not None:
+            for addr in self.sender_addrs:
+                payload += format_tlv(0x04, addr)
```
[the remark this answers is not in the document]
apaillier-ledger: The generic parser uses its own tags, it has too many structs for each property to have a unique tag.
And it is very specific to this feature of the app, and won't be used by others.

src_features/generic_tx_parser/gtp_param_trusted_name.c
```diff
@@ -87,6 +106,14 @@ bool format_param_trusted_name(const s_param_trusted_name *param, const char *na
         chain_id = get_tx_chain_id();
         for (int i = 0; i < values.size; ++i) {
             buf_shrink_expand(values.value[i].ptr, values.value[i].length, addr, sizeof(addr));
+            // replace by wallet addr if a match is found
+            for (uint8_t idx = 0; idx < param->sender_addr_count; ++idx) {
```
cedelavergne-ledger, line 110: I don't see/understand where you are doing a replacement?

Here, the loop just gets the `public_key` sereval times, but without changing any parameters.... Why would we get a different value?

apaillier-ledger: We loop and do a `memcmp` each iteration, we only do one `get_public_key` `if` the `memcmp`  is successful, and that `break`s out of the loop.

cedelavergne-ledger (approved):

PR #754, opened 2025-03-21
GCS PARAM_TOKEN formatter

apaillier-ledger:
## Description

Added new parameter type for generic clear-signing

## Changes include

- [x] Tests
- [ ] Documentation


client/src/ledger_app_clients/ethereum/tlv.py
```diff
@@ -30,27 +30,36 @@ class FieldTag(IntEnum):
...
 
 
-def der_encode(value: int) -> bytes:
-    # max() to have minimum length of 1
-    value_bytes = value.to_bytes(max(1, (value.bit_length() + 7) // 8), 'big')
-    if value >= 0x80:
-        value_bytes = (0x80 | len(value_bytes)).to_bytes(1, 'big') + value_bytes
-    return value_bytes
+class TlvSerializable:
+    def serialize(self) -> bytes:
```
cedelavergne-ledger, line 34: Why this definition?

client/src/ledger_app_clients/ethereum/tlv.py
```diff
@@ -30,27 +30,36 @@ class FieldTag(IntEnum):
...
+        assert isinstance(value, bytes), f"Unhandled TLV formatting for type : {type(value)}"
+
+        tlv = bytearray()
+        tlv += TlvSerializable.der_encode(tag)
+        tlv += TlvSerializable.der_encode(len(value))
+        tlv += value
+        return tlv
 
 
 def format_tlv(tag: int, value: Union[int, str, bytes, bytearray]) -> bytes:
```
cedelavergne-ledger, line 64: Is this needed?

client/src/ledger_app_clients/ethereum/tlv.py
apaillier-ledger: With this interface/abstract class, any other class that inherits it will have the `serialize()` method (and also the two helper static methods), and we can easily do an array of different objects that all have this method and treat them the same regardless of their actual type.

cedelavergne-ledger (approved):

PR #775, opened 2025-04-24
Update client to use Web3.py v7

apaillier-ledger:
## Description

EIP-7702 tests benefit greatly from this newer Web3.py version.

## Changes include

- [x] Tests
- [ ] Documentation

### DIFF
--- a/client/src/ledger_app_clients/ethereum/client.py
+++ b/client/src/ledger_app_clients/ethereum/client.py
@@ -219,7 +219,7 @@ class EthAppClient:
         if tx_raw is not None:
             tx = tx_raw
         else:
-            tx = Web3().eth.account.create().sign_transaction(tx_params).rawTransaction
+            tx = Web3().eth.account.create().sign_transaction(tx_params).raw_transaction
         prefix = bytes()
         suffix = []
         if tx[0] in [0x01, 0x02, 0x04]:
--- a/client/src/ledger_app_clients/ethereum/client.py
+++ b/client/src/ledger_app_clients/ethereum/client.py
@@ -213,13 +213,10 @@ class EthAppClient:
     def eip712_filtering_raw(self, name: str, sig: bytes, discarded: bool):
         return self._exchange_async(self._cmd_builder.eip712_filtering_raw(name, sig, discarded))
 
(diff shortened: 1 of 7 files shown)


client/src/ledger_app_clients/ethereum/client.py
```diff
@@ -213,13 +213,10 @@ def eip712_filtering_trusted_name(self,
     def eip712_filtering_raw(self, name: str, sig: bytes, discarded: bool):
         return self._exchange_async(self._cmd_builder.eip712_filtering_raw(name, sig, discarded))
 
-    def serialize_tx(self, tx_params: dict, tx_raw: Optional[bytes] = None) -> Tuple[bytes, bytes]:
+    def serialize_tx(self, tx_params: dict) -> tuple[bytes, bytes]:
```
cedelavergne-ledger, line 216: Why changing this? I thing the type is `Tuple` (from `typing`) and not `tuple` :thinking:

tests/ragger/test_sign.py
```diff
@@ -81,21 +77,12 @@ def common(firmware: Firmware,
             end_text = "Accept"
         else:
             end_text = r"(Sign|Accept (risk|threat))"
-        scenario_navigator.review_approve(test_name=test_name, custom_screen_text=end_text, do_comparison=test_name!="")
+        scenario_navigator.review_approve(test_name=test_name, custom_screen_text=end_text, do_comparison=test_name != "")
```
cedelavergne-ledger, line 80: Minor, but I don't think it is important to have an extra space?

client/src/ledger_app_clients/ethereum/client.py
apaillier-ledger: It seems to be deprecated : https://docs.python.org/3/library/typing.html#typing.Tuple

tests/ragger/test_sign.py
apaillier-ledger: Not strictly necessary, but Flake8 complains about it : https://www.flake8rules.com/rules/E225.html

cedelavergne-ledger (approved):

PR #787, opened 2025-05-05
Remove Nano S target

apaillier-ledger:
## Description

First step of dropping the Nano S target.
The app can still be built for Nano S, but the code will progressively break as it won't be maintained / tested for this target anymore.
No further release will be deployed for it unless really necessary.

## Changes include

- [ ] Tests
- [ ] Documentation


tdejoigny-ledger (commented):
readme + makefile + standard_plugin.mk + copy_clones.sh + asset_info.h  and other file to update (remove nanoS)

grep NanoS

PR #799, opened 2025-05-15
Adapt to ragger with Device instead of Firmware

cedelavergne-ledger:


client/src/ledger_app_clients/ethereum/eip712/InputData.py
```diff
@@ -489,24 +486,21 @@ def process_data(aclient: EthAppClient,
             pass
         prepare_filtering(filters)
 
-    if aclient._pki_client is None:
-        print(f"Ledger-PKI Not supported on '{aclient._firmware.name}'")
+    # pylint: disable=line-too-long
+    if app_client.device.type != DeviceType.NANOSP:
```
apaillier-ledger, line 490: Why change the `==` into `!=` everywhere ?

cedelavergne-ledger: :fearful: 
Thanks for your attention.
This is fixed (everywhere) :wink:

apaillier-ledger (approved):

PR #814, opened 2025-06-10
Fix test config

cedelavergne-ledger:
Add function to send the network information (in class `DynamicNetwork`)
Adapt test functions to new dynamic_networks `send_network_information`
Fix client to avoid cyclic redundency with `dynamic_networks`
Add PKI certificate to `set_external_plugin`

### DIFF
--- a/client/src/ledger_app_clients/ethereum/client.py
+++ b/client/src/ledger_app_clients/ethereum/client.py
@@ -341,6 +341,10 @@ class EthAppClient:
                             contract_address: bytes,
                             method_selelector: bytes,
                             sig: Optional[bytes] = None) -> RAPDU:
+
+        # Send ledgerPKI certificate
+        self.pki_client.send_certificate(PKIPubKeyUsage.PUBKEY_USAGE_COIN_META)
+
         if sig is None:
             # Temporarily get a command with an empty signature to extract the payload and
             # compute the signature on it


client/src/ledger_app_clients/ethereum/dynamic_networks.py
```diff
@@ -0,0 +1,40 @@
...
+                 chain_id: Optional[int],
+                 icon: Optional[bytes] = None) -> None:
+        self.name = name
+        self.ticker = ticker
+        self.chain_id = chain_id
+        self.icon = icon
+
+    def serialize(self) -> bytes:
+        assert self.chain_id is not None, "Chain_id required"
+        assert self.name, "Name is required"
```
apaillier-ledger, line 26: These checks seem redundant.

client/src/ledger_app_clients/ethereum/dynamic_networks.py
```diff
@@ -0,0 +1,40 @@
...
+        assert self.ticker, "Ticker is required"
+        # Construct the TLV payload
+        payload: bytes = self.serialize_field(FieldTag.STRUCT_TYPE, 8)
+        payload += self.serialize_field(FieldTag.STRUCT_VERSION, 1)
+        payload += self.serialize_field(FieldTag.BLOCKCHAIN_FAMILY, 1)
+        payload += self.serialize_field(FieldTag.CHAIN_ID, self.chain_id.to_bytes(8, 'big'))
```
apaillier-ledger, line 32: The `chain_id` could be passed as is (an `int`), so that it would benefit from the TLV serialization and be stored as up to 8 bytes depending on its value, and not necessarily as 8 bytes all the time.

cedelavergne-ledger: Fixed

client/src/ledger_app_clients/ethereum/dynamic_networks.py
```diff
@@ -0,0 +1,40 @@
...
+from hashlib import sha256
+
+from .tlv import TlvSerializable, FieldTag
+from .keychain import sign_data, Key
+
+
+class DynamicNetwork(TlvSerializable):
+    name: str
+    ticker: str
+    chain_id: Optional[int]
```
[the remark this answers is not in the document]
apaillier-ledger: It does not make sense to try to instantiate a `DynamicNetwork` without a `chain_id`, especially since there is a check for it in `provide_network_information` so that it does not send it in this case.

client/src/ledger_app_clients/ethereum/dynamic_networks.py
cedelavergne-ledger: Removed

PR #818, opened 2025-06-13
Proxy info - spec alignment

apaillier-ledger:
## Description

Spec has changed since the initial implementation.

## Changes include

- [x] Tests
- [ ] Documentation


src_features/provide_proxy_info/proxy_info.c
```diff
@@ -159,12 +172,20 @@ bool verify_proxy_info_struct(const s_proxy_info_ctx *context) {
         PRINTF("Error: could not finalize struct hash!\n");
         return false;
     }
+    if (context->struct_type != TYPE_PROXY_INFO) {
```
cedelavergne-ledger, line 175: Why not checking this value directly in `handle_type()`?

apaillier-ledger: To separate data parsing from data handling.

PR #822, opened 2025-06-19
EIP-712 memory allocator migration

apaillier-ledger:
## Description

Follow up to #792

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/list.c
+++ b/src/list.c
@@ -1,0 +1,131 @@
+#include "list.h"
+
+void flist_push_front(s_flist_node **list, s_flist_node *node) {
+    if ((list != NULL) && (node != NULL)) {
+        node->next = *list;
+        *list = node;
+    }
+}
+
+void flist_pop_front(s_flist_node **list, f_list_node_del func) {
+    s_flist_node *tmp;
+
+    if (list != NULL) {
+        tmp = *list;
+        if (tmp != NULL) {
+            *list = tmp->next;
+            if (func != NULL) func(tmp);
+        }
+    }
+}
+
+void flist_push_back(s_flist_node **list, s_flist_node *node) {
+    s_flist_node *tmp;
+
+    if ((list != NULL) && (node != NULL)) {
+        if (*list == NULL) {
+            *list = node;
+        } else {
+            tmp = *list;
+            if (tmp != NULL) {
+                for (; tmp->next != NULL; tmp = tmp->next)
+                    ;
+                tmp->next = node;
+            }
+        }
+    }
+}
+
+void flist_pop_back(s_flist_node **list, f_list_node_del func) {
+    s_flist_node *tmp;
+
+    if (list != NULL) {
+        tmp = *list;
+        if (tmp != NULL) {
+            // only one element
+            if (tmp->next == NULL) {
+                flist_pop_front(list, func);
+            } else {
+                for (; tmp->next->next != NULL; tmp = tmp->next)
+                    ;
+                if (func != NULL) func(tmp->next);
+                tmp->next = NULL;
+            }
+        }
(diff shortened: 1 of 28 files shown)


src/list.c
```diff
@@ -0,0 +1,131 @@
...
+    if ((list != NULL) && (node != NULL)) {
+        node->next = *list;
+        *list = node;
+    }
+}
+
+void flist_pop_front(s_flist_node **list, f_list_node_del func) {
+    s_flist_node *tmp;
+
+    if (list != NULL) {
```
cedelavergne-ledger, line 13: it could be also 
`if ((list != NULL) && (*list != NULL)) {`

src/list.c
```diff
@@ -0,0 +1,131 @@
...
+        tmp = *list;
+        if (tmp != NULL) {
+            *list = tmp->next;
+            if (func != NULL) func(tmp);
```
cedelavergne-ledger, line 17: to be checked at the begining of the function?
If Null, should probably return an error.....

cedelavergne-ledger:
Despites everyone don't agree, I am really convinced the `src/list.c` should be moved to the SDK to be global, efficient, robust, and also deeply tested with unit-tests and fuzzing :wink:

src/list.c
apaillier-ledger: Not necessarily, it would still pop it from the list, but it would not be freed which might be needed sometimes.

src_features/signMessageEIP712/ui_logic.c
```diff
@@ -13,18 +13,20 @@
...
-typedef struct {
-    // display name, not NULL-terminated
-    char name[25];
-    uint8_t name_length;
-    uint8_t value[INT256_LENGTH];
-    uint8_t value_length;
+typedef struct amount_join {
+    s_flist_node _list;
+    // display name, NULL-terminated
+    char name[25 + 1];
```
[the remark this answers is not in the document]
apaillier-ledger: Good catch, thanks!

src_features/signMessageEIP712/ui_logic.c
```diff
@@ -52,19 +59,31 @@ typedef struct {
...
 } t_ui_context;
 
 static t_ui_context *ui_ctx = NULL;
 
+static void delete_filter_crc(s_filter_crc *fcrc) {
+    app_mem_free(fcrc);
+}
+
+static void delete_ui_pair(s_ui_712_pair *pair) {
+    if (pair->key != NULL) app_mem_free(pair->key);
```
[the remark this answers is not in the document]
apaillier-ledger: `s_ui_712_pair` and other structs like it are linked-list nodes.
These `delete_xxx` functions are only used as callbacks from linked-list functions, such as : 
* `flist_pop_front`
* `flist_pop_back`
* `flist_remove`
* `flist_clear`
These functions handle the `NULL` checks before calling their callback and set the appropriate pointer(s) to `NULL` afterwards.

src/list.c
cedelavergne-ledger: Oh, so we could have a memory leak here, because if `func` is NULL, we'll point on the 2nd element (`tmp->next`), and the old one (`tmp`) will not be freed, and then, be lost :thinking:

apaillier-ledger: It could (if the node was dynamically allocated).
But it's not an error for the linked-list, and does not prevent it from working.
And it also allows us to pop a node from one list and add it to another without having to allocate a duplicate node and copy the old one.

cedelavergne-ledger (approved):
Still disagree with some answers... But let's go like this for the moment.

PR #850, opened 2025-08-22
RLP handling fix

apaillier-ledger:
## Description

If a `SIGN_TX` APDU chunk ended perfectly just after the access list, it could consider the RLP parsing to be done, meanwhile the wallet would still send the (usually empty) y_parity, r, s in a final `SIGN_TX` APDU that the app would refuse.

## Changes include

- [ ] Tests
- [ ] Documentation


apaillier-ledger:
Not worth it, ETH signer will enforce the removal of these fields instead.

PR #863, opened 2025-09-18
[Security Audit] Fix 1.19.0

cedelavergne-ledger:
## Description

Fix Audit findings from https://github.com/LedgerHQ/app-ethereum/pull/862

Bonus:
- Update the workflows to align on reusable ones
- Update ioTex icons

## Changes include

- [ ] Tests
- [ ] Documentation


bboilot-ledger:
Pushed some improvement for the fuzzer. However, since we switched to the dynamic allocator, the fuzzer is not able to follow allocations:
```
INFO: libFuzzer disabled leak detection after every mutation.
      Most likely the target function accumulates allocated
      memory in a global state w/o actually leaking it.
      You may try running this binary with -trace_malloc=[12]      to get a trace of mallocs and frees.
      If LeakSanitizer is enabled in this process it will still
      run on the process shutdown.
```

I think that I am incorrectly instantiating/cleaning the app's feature.... :/ That's not mandatory for the review since we have valground!

bboilot-ledger (dismissed):

PR #869, opened 2025-09-26
Add Sei Network

besated:
## Description

Add Sei to the supported networks so the correct ticker and icon appear when interacting with this chain

## Changes include

- [ ] Tests
- [ ] Documentation


apaillier-ledger:
Done in #954

PR #960, opened 2026-03-05
Fix medium bugs

cedelavergne-ledger:
## Description

https://ledgerhq.atlassian.net/browse/B2CA-2562

## Changes include

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/features/generic_tx_parser/gtp_data_path.c
+++ b/src/features/generic_tx_parser/gtp_data_path.c
@@ -129,7 +129,7 @@ static bool path_leaf(const s_leaf_args *leaf,
     uint8_t *leaf_buf = NULL;
     uint8_t cpy_length;
 
-    if (collection->size > MAX_VALUE_COLLECTION_SIZE) {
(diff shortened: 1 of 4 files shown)


src/features/generic_tx_parser/gtp_data_path.c
```diff
@@ -245,15 +247,23 @@ static bool path_array(const s_array_args *array,
         end = array_size;
     }
 
+    passes = end - start;
+    if (passes > UINT8_MAX) {
```
bboilot-ledger: Seems overcomplicated. Wouldn't be easier to just check `end < start` ? We can still keep passes variable if we indeed want to avoid reusing `(end -start)` at two different places.

bboilot-ledger (commented):
Thanks for the work 🙏

src/features/generic_tx_parser/gtp_data_path.c
cedelavergne-ledger: Fixed

apaillier-ledger (approved):

PR #990, opened 2026-04-13
B2CA-2548: Address book

cedelavergne-ledger:
## Description

- Update client to support Address Book
- Add Address Book feature
- Add Address Book tests

## Changes include

- [ ] Tests
- [ ] Documentation


Makefile
```diff
@@ -159,13 +158,18 @@ ENABLE_NBGL_QRCODE = 1
...
 ########################################
+ENABLE_NBGL_FOR_NANO_DEVICES = 1
 ENABLE_PKI_LIBRARY = 1
 ENABLE_DYNAMIC_ALLOC = 1
 ENABLE_TLV_LIBRARY = 1
 ENABLE_LISTS_LIBRARY = 1
+ifeq ($(TARGET_NAME),$(filter $(TARGET_NAME),TARGET_STAX TARGET_FLEX TARGET_APEX_M TARGET_APEX_P))
+    ENABLE_ADDRESS_BOOK = 1
+    ENABLE_ADDRESS_BOOK_LEDGER_ACCOUNT = 1
+endif
```
fbeutin-ledger: Do we want this enabled on all chains?

src/features/address_book/handle_provide_contact.c
```diff
@@ -0,0 +1,128 @@
...
+        return false;
+    }
+    memset(node, 0, sizeof(*node));
+
+    strncpy(node->contact_name, contact->contact_name, sizeof(node->contact_name) - 1);
+    strncpy(node->scope, contact->scope, sizeof(node->scope) - 1);
+    memcpy(node->identifier, contact->identifier, ADDRESS_LENGTH);
+    node->chain_id = contact->chain_id;
+
+    flist_push_back((flist_node_t **) &g_contact_list, (flist_node_t *) node);
```
fbeutin-ledger: What happens if the LW sends duplicates?

Makefile
cedelavergne-ledger: Not mandatory. It is up to each App to select if it has to be supported, and then for which config/chain.
Here, it allows to demonstrate the feature.
But for Ethereum, I guess it will be for all chains (to be confirmed by the Product)

tests/functional/test_address_book.py
```diff
@@ -0,0 +1,656 @@
+import hashlib
```
fbeutin-ledger, line 3: Two major remarks on this :

- We should move a lot of this into ragger into a helper client for address book. Otherwise, all applications will have to redefine a lot of this test stuff (just like what happened with PKI). OK for merge now, we can delay the move to Ragger for the first non ETH user

- Tests are a bit light on API solidity:
- Reject paths: invalid chain_id, wrong blockchain family, wrong identifier length, wrong HMAC, wrong group_handle, user-rejection at UI review. Each should be one test.
- Chunk framing: truncated first chunk (< 2-byte header), P2=0x80 without a prior P2=0x00, announced length larger than actual data, interleaved P1 chunks (see §3.2 above).
- Contact lookup: provide_contact followed by sign TX to a different address (should display raw address, not substitute), and to the same address with different chain_id (should not substitute).
- Legacy-device negative test: build for nanos2/nanox and confirm the feature is cleanly disabled (no dead symbols).

src/features/address_book/handle_provide_contact.c
cedelavergne-ledger: It is a corner case... With linked list, it will not be an _error_, because the duplicates will be appended to the end of the list, and for the Tx itself, we'll stop on the 1st element found.
Rememebr that the PR is is not yet complete because I have focused on the standard TX only, and other types of TX will have to handle the Address Book.
Then, I will also address this corner case to drop the duplicate (ro send back a dedicated error code... To be defined with Product)

src/features/address_book/handle_identity.c
```diff
@@ -0,0 +1,387 @@
...
+        return false;
+    }
+
+    // Allocate contact name
+    g_ctx.edit_identifier.contact_name = APP_MEM_ALLOC(CONTACT_NAME_LENGTH);
+    if (g_ctx.edit_identifier.contact_name == NULL) {
+        PRINTF("Failed to allocate contact_name\n");
+        return false;
+    }
+    strncpy(g_ctx.edit_identifier.contact_name,
```
fbeutin-ledger: strlcpy

src/features/address_book/handle_identity.c
```diff
@@ -0,0 +1,387 @@
...
+            CONTACT_NAME_LENGTH - 1);
+
+    // Allocate scope
+    g_ctx.edit_identifier.scope = APP_MEM_ALLOC(SCOPE_LENGTH);
+    if (g_ctx.edit_identifier.scope == NULL) {
+        PRINTF("Failed to allocate scope\n");
+        APP_MEM_FREE_AND_NULL((void **) &g_ctx.edit_identifier.contact_name);
+        return false;
+    }
+    strncpy(g_ctx.edit_identifier.scope, params->identity.scope, SCOPE_LENGTH - 1);
```
fbeutin-ledger: strlcpy

src/features/address_book/handle_provide_contact.c
```diff
@@ -0,0 +1,128 @@
...
+
+    node = APP_MEM_ALLOC(sizeof(*node));
+    if (node == NULL) {
+        PRINTF("handle_provide_contact: allocation failed\n");
+        return false;
+    }
+    memset(node, 0, sizeof(*node));
+
+    strncpy(node->contact_name, contact->contact_name, sizeof(node->contact_name) - 1);
+    strncpy(node->scope, contact->scope, sizeof(node->scope) - 1);
```
fbeutin-ledger: strlcpy

src/features/address_book/handle_provide_contact.c
```diff
@@ -0,0 +1,128 @@
...
+/* Exported functions --------------------------------------------------------*/
+
+/**
+ * @brief Release all stored Address Book contacts.
+ *
+ * Iterates the linked list and frees every node. Should be called when
+ * the application returns to idle so that the next signing flow starts
+ * with an empty contact list.
+ */
+void address_book_contact_cleanup(void) {
```
fbeutin-ledger: I don't see it ever called?

tests/functional/test_address_book.py
cedelavergne-ledger: Agree to put lots of common stuff into ragger. But As you said, it is important to validate the principle 1st, and then we'll prepare an update of the ragger to align all Apps

Will look at adding more tests

The contact lookup should be ok. when a Tx is received, we check for the address to replace by a name. If it is not found we keep the raw address (same as Tructed Name existing feature)

For the Legacy devices, yes, this is fixed

src/features/address_book/handle_provide_contact.c
cedelavergne-ledger: Agree, not yet called, because the received Address Book contacts may live during the whole App session... This is to be confirmed, and thus I prefer to keep the function for now.

src/features/address_book/handle_identity.c
cedelavergne-ledger: Fixed

src/features/address_book/handle_provide_contact.c
cedelavergne-ledger: Fixed

src/features/address_book/handle_identity.c
cedelavergne-ledger: Fixed

src/features/address_book/handle_provide_contact.c
cedelavergne-ledger: Fixed: Duplicate entry is simply ignored

tests/functional/test_address_book.py
cedelavergne-ledger: Status:
- Reject paths:
  - invalid chain_id -> Tested with Unit-Tests
  - wrong blockchain family -> Tested with Unit-Tests
  - wrong identifier length -> Covered by lib_tlv
  -  wrong HMAC -> Will be done in SDK after the OS rework
  -  wrong group_handle -> Will be done in SDK after the OS rework
  - user-rejection at UI review -> Done with ragger
- Chunk framing:
  - truncated first chunk (< 2-byte header) -> Will be done in SDK after the OS rework
  - P2=0x80 without a prior P2=0x00 -> Will be done in SDK after the OS rework
  - announced length larger than actual data -> Will be done in SDK after the OS rework
  - interleaved P1 chunks -> Will be done in SDK after the OS rework
- Contact lookup:
  - provide_contact followed by sign TX to a different address -> Fallback already existing with Trusted Names
  - to the same address with different chain_id ->  -> Done with ragger

client/src/ledger_app_clients/ethereum/address_book.py
```diff
@@ -0,0 +1,495 @@
...
+    """
+    _CLA: int = 0xB0
+    _INS: int = 0x10
+
+    def __init__(self, backend: BackendInterface) -> None:
+        """Initialize the Address Book client with a backend."""
+        self._backend = backend
+
+    def send_async_raw(self, p1: AddressBookSubCommand, payload: bytes):
+        """Send an address book APDU, splitting into chunks when needed.
```
apaillier-ledger: This APDU should be added to doc/ethapp.adoc

client/src/ledger_app_clients/ethereum/address_book.py
```diff
@@ -0,0 +1,495 @@
...
+        transported as multiple APDUs using a P2-based chunking scheme:
+          - First chunk  (P2=0x00): 2-byte big-endian total length + first slice
+          - Next  chunks (P2=0x80): subsequent slices
+        All intermediate chunks are sent synchronously (device responds 9000).
+        The last chunk is sent asynchronously to allow the UI flow to proceed.
+
+        All other commands fit in a single short APDU (P2=0x00, no framing).
+        """
+        MAX_CHUNK = 255
+        P2_FIRST = 0x00
```
apaillier-ledger: Why does this follow the 0x00 / 0x80 scheme that the SIGN_TX (0x04) and SIGN_MESSAGE (0x08) APDUs have instead of the 0x01 / 0x00 that all other APDUs carrying TLV have ?

client/src/ledger_app_clients/ethereum/address_book.py
```diff
@@ -0,0 +1,495 @@
...
+                      AddressBookSubCommand.SUB_CMD_PROVIDE_LEDGER_ACCOUNT_CONTACT):
+            header = bytes([self._CLA, self._INS, p1, P2_FIRST, len(payload)])
+            return self._backend.exchange_async_raw(header + payload)
+
+        # Prepend 2-byte big-endian total length (mirrors the firmware framing).
+        framed = len(payload).to_bytes(2, 'big') + payload
+        chunks = [framed[i:i + MAX_CHUNK] for i in range(0, len(framed), MAX_CHUNK)]
+
+        # Send all but the last chunk synchronously.
+        for i, chunk in enumerate(chunks[:-1]):
```
apaillier-ledger: The APDU encapsulation should not be part of this Python class, this should only handle the TLV serialization like other classes do, since that's what's really common to all apps supporting the Address Book. The APDU can and probably will be app-specific.

apaillier-ledger (changes requested):

client/src/ledger_app_clients/ethereum/address_book.py
apaillier-ledger: Nevermind, it's a `0xB0` APDU, not processed by the app directly.

client/src/ledger_app_clients/ethereum/address_book.py
apaillier-ledger: Nevermind, it's a `0xB0` APDU, not processed by the app directly, so it's not necessary to be aligned with the app's APDUs.

client/src/ledger_app_clients/ethereum/address_book.py
cedelavergne-ledger: Refactoring done

client/src/ledger_app_clients/ethereum/address_book.py
```diff
@@ -0,0 +1,431 @@
...
+SCOPE_MAX_LENGTH = 32         # SCOPE_LENGTH(33) - 1 null terminator
+DEFAULT_BIP32_PATH = "m/44'/60'/0'/0/0"
+DEFAULT_CONTACT_NAME = "Alice"
+DEFAULT_CHAIN_ID = 1  # Ethereum Mainnet
+DEFAULT_SCOPE = "Eth Address 1"
+DEFAULT_ADDRESS = bytes.fromhex("6b175474e89094c44da98b954eedeac495271d0f")
+DEFAULT_ACCOUNT_NAME = "ETH main address"
+
+
+class AddressBookClient(TlvSerializable):
```
apaillier-ledger: Maybe this should be split into 8 distinct classes, one for each subcommand (all derived from `TlvSerializable`).
So that they could be handled transparently as any other `TlvSerializable` derived objects, because as it is now, calling `.serialize()` would raise a `NotImplementedError`.

apaillier-ledger (changes requested):

client/src/ledger_app_clients/ethereum/address_book.py
cedelavergne-ledger: ok, improved

client/src/ledger_app_clients/ethereum/command_builder.py
```diff
@@ -431,6 +437,28 @@ def provide_trusted_name(self, tlv_payload: bytes) -> list[bytes]:
...
+        Every sub-command uses the same chunked transport: the payload is
+        prefixed with its 2-byte big-endian total length and split over
+        255-byte chunks (a payload fitting in one chunk still carries the
+        prefix). The firmware reassembles it in reassemble_chunks().
+        """
+        payload = struct.pack(">H", len(tlv_payload)) + tlv_payload
+
+        chunks = []
+        p2 = self._AB_P2_FIRST_CHUNK
+        while len(payload) > 0:
```
apaillier-ledger: Could this use `common_tlv_serialize` ?

cedelavergne-ledger: Refactored and improved

apaillier-ledger (approved):

PR #1031, opened 2026-05-12
Fix secu review

### LINKED ISSUE #988: [High] Stale schema hash lets EIP-712 metadata signatures apply to a different schema
## Location

- **File**: `src/features/sign_message_eip712/commands_712.c`
- **Lines**: 204-208

## Classification

| Property | Value |
| --- | --- |
| **Severity** | High |
| **Category** | Other |
| **CWE** | CWE-CWE-345 |
| **OWASP** | A08:2021-Software and Data Integrity Failures |
| **Confidence** | Suspected |
| **Likelihood** | Medium |

## Technical Description
The trusted-filtering design binds Ledger-signed metadata to `eip712_context->schema_hash`, but the schema hash is computed as soon as `P2_FILT_ACTIVATE` is received, before the type system is locked. `handle_eip712_struct_def()` still accepts additional struct names and fields until `struct_state` becomes `DEFINED`, and that only happens later in `path_set_root()` during struct implementation. Because `sig_verif_start()` always verifies filtering signatures against the previously stored `schema_hash`, a malicious host can: (1) define a benign schema, (2) call `P2_FILT_ACTIVATE` to store its hash, (3) append extra fields/structs to create a different final schema, then (4) reuse valid Ledger-signed filter metadata for the old schema. The device will verify the metadata against the stale hash while hashing/signing the modified typed data, so extra attacker-controlled fields can be hidden from the on-device review.

## Vulnerable Code

```c
        case P2_FILT_ACTIVATE:
            if (!N_storage.verbose_eip712) {
                ui_712_set_filtering_mode(EIP712_FILTERING_FULL);
                ret = compute_schema_hash();
            }
            forget_known_assets();
            break;
```

## Impact
USB/Bluetooth/BLE: attacker controlling the connected host application or dApp can compute the trusted schema hash over a benign EIP-712 schema, mutate the schema afterward, and still pass Ledger-signed filtering metadata verification. This can cause the device to present a trusted filtered review for one message type while signing a different typed-data payload containing hidden fields, enabling deceptive permits/approvals or other unauthorized authorizations.

## Remediation
Lock the schema before computing or trusting `schema_hash`. Either reject `P2_FILT_ACTIVATE` until all type definitions are finalized, or recompute the schema hash immediately before every metadata-signature verification and before signing, and abort if any struct definition changed after activation.

---
*Created by [Cerberus](https://github.com/Donjon-Cerberus) Merlin*

### LINKED ISSUE #993: [High] Date/time review truncates multi-byte uint values in ui_712_format_datetime
## Location

- **File**: `src/features/sign_message_eip712/ui_logic.c`
- **Lines**: 708-719

## Classification

| Property | Value |
| --- | --- |
| **Severity** | High |
| **Category** | Other |
| **CWE** | CWE-CWE-451 |
| **OWASP** | A04:2021-Insecure Design |
| **Confidence** | Confirmed |
| **Likelihood** | Medium |

## Technical Description
The date/time formatter accepts any `uint` field selected by a trusted date-time filter, but it converts the raw value with `u64_from_BE(data, length)`. `u64_from_BE()` only consumes the first 8 bytes of the supplied buffer. For EIP-712 `uint256` timestamps, or for non-minimal encodings with attacker-added leading zero bytes, the device displays a truncated/smaller timestamp while the signing hash is computed from the full integer value. Because leading zero bytes do not change the final 32-byte EIP-712 integer encoding, a malicious host can keep the signed value valid while showing a much earlier date on-screen.

## Vulnerable Code

```c
static bool ui_712_format_datetime(const uint8_t *data,
                                   uint8_t length,
                                   const s_struct_712_field *field_ptr) {
    time_t timestamp;

    if ((length >= field_ptr->type_size) && ismaxint((uint8_t *) data, length)) {
        snprintf(strings.tmp.tmp, sizeof(strings.tmp.tmp), "Unlimited");
        return true;
    }
    timestamp = u64_from_BE(data, length);
    return time_format_to_utc(&timestamp, strings.tmp.tmp, sizeof(strings.tmp.tmp));
}
```

## Impact
Network: a remote wallet-connected attacker can craft EIP-712 timestamp/deadline fields so the review screen shows an earlier or expired-looking date while the signed message contains a longer-lived deadline. This can trick the user into approving signatures that remain valid far longer than displayed.

## Remediation
Only allow date/time formatting for integers that fit within 64 bits, and reject non-minimal encodings. Alternatively, normalize the value first by stripping leading zeros and then require the remaining length to be <= 8 before calling `u64_from_BE()`.

---
*Created by [Cerberus](https://github.com/Donjon-Cerberus) Merlin*

### LINKED ISSUE #994: [High] Stale-context UI spoofing in ERC20 plugin when calldata is incomplete or non-conforming
## Location

- **File**: `src/plugins/erc20/erc20_plugin.c`
- **Lines**: 35-163

## Classification

| Property | Value |
| --- | --- |
| **Severity** | High |
| **Category** | Other |
| **CWE** | CWE-CWE-451 |
| **OWASP** | A04:2021-Insecure Design |
| **Confidence** | Suspected |
| **Likelihood** | Medium |

## Technical Description
The ERC20 internal plugin reuses the shared plugin context without clearing sensitive display fields (`destinationAddress`, `amount`, `ticker`, `decimals`, `selectorIndex`) and never verifies that the required calldata words were actually received before returning `ETH_PLUGIN_RESULT_OK`. Only `extra_data` is zeroed during init.

As a result, if a new transaction provides incomplete or non-conforming calldata, the plugin can still finalize successfully and render a trusted token UI using stale values left over from a previous signing flow. The parameter parser only fills fields when specific offsets are encountered, and finalize does not require that both the address and amount were parsed. In the signing flow, a successful plugin result suppresses the generic calldata/blind-signing warning and treats the plugin UI as authoritative.

This is especially dangerous for known-token contract addresses (where token metadata is available and the ERC20 plugin remains on the trusted UI path). Many ERC20 contracts are proxy/upgradeable or use custom dispatch logic, so the plugin must not assume that a selector match alone guarantees standard ABI semantics.

## Vulnerable Code

```c
case ETH_PLUGIN_INIT_CONTRACT: {
    ethPluginInitContract_t *msg = (ethPluginInitContract_t *) parameters;
    erc20_parameters_t *context = (erc20_parameters_t *) msg->pluginContext;

    explicit_bzero(context->extra_data, sizeof(context->extra_data));
    context->extra_data_len = 0;

    // enforce that ETH amount should be 0
    if (!allzeroes(msg->txContent->value.value, CALLDATA_CHUNK_SIZE)) {
        msg->result = ETH_PLUGIN_RESULT_ERROR;
    } else {
        size_t i;
        for (i = 0; i < NUM_ERC20_SELECTORS; i++) {
            if (memcmp((uint8_t *) PIC(ERC20_SELECTORS[i]),
                       msg->selector,
                       CALLDATA_SELECTOR_SIZE) == 0) {
                context->selectorIndex = i;
                break;
            }
        }
        ...
        msg->result = ETH_PLUGIN_RESULT_OK;
    }
} break;
...
case ETH_PLUGIN_PROVIDE_PARAMETER: {
    ...
    switch (msg->parameterOffset) {
        case CALLDATA_SELECTOR_SIZE:
            memmove(context->destinationAddress,
                    msg->parameter + (CALLDATA_CHUNK_SIZE - ADDRESS_LENGTH),
                    ADDRESS_LENGTH);
            msg->result = ETH_PLUGIN_RESULT_OK;
            break;
        case CALLDATA_SELECTOR_SIZE + CALLDATA_CHUNK_SIZE:
            memmove(context->amount, msg->parameter, CALLDATA_CHUNK_SIZE);
            msg->result = ETH_PLUGIN_RESULT_OK;
            break;
        default:
            ...
    }
} break;
...
case ETH_PLUGIN_FINALIZE: {
    ...
    msg->tokenLookup1 = msg->txContent->destination;
    msg->numScreens = 2;
    if (context->extra_data_len > 0) {
        msg->numScreens += 1;
    }
    ...
    msg->uiType = ETH_UI_TYPE_GENERIC;
    msg->result = ETH_PLUGIN_RESULT_OK;
} break;
```

## Impact
Network: a remote attacker controlling the host/dApp can send malformed or selector-compatible calldata to a known token contract and obtain a trusted Ledger ERC20 confirmation screen that shows stale recipient/amount values instead of the real call data. This can mislead the user into signing an unintended token operation while bypassing the normal generic-data warning path. Blast radius includes supported ERC20 token contracts, particularly upgradeable/proxy tokens whose runtime behavior can diverge from the plugin’s fixed assumptions.

## Remediation
Zero the full ERC20 plugin context during init, and fail finalize unless both mandatory parameters were parsed for the selected method. Track parse completion explicitly (for example, booleans for `seen_destination` and `seen_amount`, or a state machine that must reach a terminal state) and reject unexpected calldata shapes instead of defaulting to success. Clearing `dataContext.tokenContext.pluginContext` on app reset is also recommended to prevent cross-transaction state reuse.

---
*Created by [Cerberus](https://github.com/Donjon-Cerberus) Merlin*

cedelavergne-ledger:
## Description

https://ledgerhq.atlassian.net/browse/B2CA-2627

- Addressing the file SECURITY_REVIEW_FULL.md
- Closes https://github.com/LedgerHQ/app-ethereum/issues/988
- Closes https://github.com/LedgerHQ/app-ethereum/issues/993
- Closes https://github.com/LedgerHQ/app-ethereum/issues/994

## Changes include

- [ ] Tests
- [ ] Documentation


src/nbgl/ui_utils.c
```diff
@@ -49,7 +49,7 @@ bool ui_pairs_init(uint8_t nbPairs) {
     }
 
     // Allocate the pairs memory
-    if (!APP_MEM_CALLOC((void **) &g_pairs, nbPairs * sizeof(nbgl_contentTagValueList_t))) {
+    if (!APP_MEM_CALLOC((void **) &g_pairs, nbPairs * sizeof(nbgl_contentTagValue_t))) {
```
apaillier-ledger: `sizeof(*g_pairs)` would avoid these typing issues.

src/features/provide_trusted_name/trusted_name.c
```diff
@@ -411,7 +411,13 @@ static bool handle_owner_deriv_path(const tlv_data_t *data, s_trusted_name_ctx *
         PRINTF("OWNER: failed to extract\n");
         return false;
     }
-    context->owner_deriv_path.length = field.ptr[0];
+    // Validate the host-supplied length against the fixed-size path buffer before any assignment.
+    uint8_t deriv_length = field.ptr[0];
```
apaillier-ledger: Redundant with the checks done in `bip32_path_read`.

src/features/sign_message_eip712/ui_logic.c
```diff
@@ -639,6 +639,22 @@ static bool ui_712_format_datetime(const uint8_t *data,
...
         return true;
     }
+    // u64_from_BE() reads from the front of the buffer; for a wide EIP-712
+    // type (e.g. uint256), the trailing 8 bytes carry the value and the
+    // leading bytes are zero-padding. Strip the padding before decoding so
+    // the displayed timestamp matches what is being hashed, and refuse the
+    // value entirely if it doesn't fit in a uint64.
+    if (length > sizeof(uint64_t)) {
+        uint8_t leading = length - sizeof(uint64_t);
+        for (uint8_t i = 0; i < leading; ++i) {
```
apaillier-ledger: Can this call `is_zeroes_buffer` instead ?

apaillier-ledger (changes requested):

src/nbgl/ui_utils.c
cedelavergne-ledger: Ok. Fixed

src/features/sign_message_eip712/ui_logic.c
cedelavergne-ledger: Good catch :wink: 
Updated

src/features/provide_trusted_name/trusted_name.c
cedelavergne-ledger: Removed

apaillier-ledger (approved):

PR #1059, opened 2026-06-15
Add Absolution-based fuzzing

N3ur0sis:
## Description

This PR integrates the SDK's Absolution-based, state-aware fuzzing framework into the Ethereum app, the app ships three targets: fuzz_app drives a sequence of APDUs through the real apdu_parser() → handleApdu() path; fuzz_plugin drives one internal plugin through the production eth_plugin_call() sequence; fuzz_parser drives the generic tx parser, EIP-712 and the calldata store at their own entry points. The latter two exist because the APDU path structurally cannot reach that code. All three go through the framework contract, so they inherit the prefix-aware mutator and the lane split.

Input is [ prefix | tail ]: the prefix restores globals to a target state via the declarative Absolution model, the tail drives the harness. Sequences matter because tlv_from_apdu() accumulates a descriptor across APDUs and one APDU carries at most 253 bytes — a single dispatch can never complete a descriptor carrying a signature.

## Changes include

- [x] Tests
- [x] Documentation


tests/fuzzing/mock/mocks.c
```diff
@@ -0,0 +1,268 @@
...
+void app_main(void) {
+}
+
+void app_quit(void) {
+}
+
+/* Faithful mirror of src/main.c's reset_app_context (main.c is excluded from
+ * the fuzz build). Handlers call this mid-dispatch (e.g. INS_SIGN P1_FIRST), so
+ * it must clear the same cross-operation state the device does; keep in sync. */
+void reset_app_context(void) {
```
bboilot-ledger: what modificiation would we need to avoid mocking the main.c and reuse its code instead?

tests/fuzzing/src/fuzz_handle_apdu.c
```diff
@@ -0,0 +1,171 @@
+/*
+ * Fuzz-local mirror of src/main.c's `static handleApdu()`.
```
bboilot-ledger: would prefer updating main.c

tests/fuzzing/src/fuzz_command_registry.inc
```diff
@@ -0,0 +1,42 @@
...
+ * Canonical fuzz command registry.
+ *
+ * Field order:
+ *   group, ins_symbol, p1_max, p2_max, flags, payload_kind, tlv_kind, allow_empty_data
+ *
+ * Keep existing command indices stable for corpus compatibility. New production
+ * APDU entries should be appended unless an ABI break and harness_version bump
+ * are intentional.
+ */
+FUZZ_COMMAND(APDU, INS_GET_PUBLIC_KEY,                1,    0,    FUZZ_CMD_HAS_DATA, RAW, NONE,            false)
```
bboilot-ledger: I thought we could directly fuzz these commands from the dispatcher. It shouldn't be needed. 
We should only fuzz dispatcher, internal plugins and swap, so it should be more than 3 harnesses.

bboilot-ledger (changes requested):

src/features/sign_message/cmd_sign_message.c
```diff
@@ -61,7 +61,11 @@ static uint16_t first_apdu_data(uint8_t **data, uint8_t *length) {
     }
 
     // Get the message length
-    signMsgCtx->msg_length = U4BE(*data, 0);
+    uint32_t msg_length = U4BE(*data, 0);
+    if (msg_length > UINT16_MAX) {
+        return SWO_INCORRECT_DATA;
+    }
+    signMsgCtx->msg_length = (uint16_t) msg_length;
```
cedelavergne-ledger: I would prefer to keep the original
`signMsgCtx->msg_length = U4BE(*data, 0);`
Why did you change this?

src/handle_apdu.c
```diff
@@ -0,0 +1,254 @@
...
+#include "cmd_trusted_name.h"
+#include "cmd_tx_info.h"
+#include "commands_712.h"
+#include "commands_7702.h"
+
+#ifdef HAVE_ETH2
+#include "withdrawal_index.h"
+#endif
+
+void reset_app_context(void) {
```
cedelavergne-ledger: Not sure this function should be located in a file called `handle_apdu`.
Then, I would suggest to add a new `app_context.c`? We could also move `app_quit` in this latter?

Same for below `pasreBip32`. It could be moved to the new file?

tests/fuzzing/mock/mocks.h
```diff
@@ -1,6 +1,20 @@
...
 #include "exceptions.h"
+#include <stdio.h>
+#include <stdint.h>
+
+#include "fuzz_defs.h"
 
 extern try_context_t fuzz_exit_jump_ctx;
+
+#define FUZZ_CTRL_SIZE 16
+extern uint8_t fuzz_ctrl[FUZZ_CTRL_SIZE];
```
cedelavergne-ledger: Such variable should be explained somewhere
Why is it declared?
What is it used for?
What does it contain?

Same for pointer below...

tests/fuzzing/invariants/domain-overrides.txt
cedelavergne-ledger, line 1: Not clear who/How to update this file
Or create it from scratch for new Apps...

tests/fuzzing/scripts/fuzz_command_registry.py
cedelavergne-ledger, line 1: For all the scripts, who is expecting to use them? To update them?
What is their goal?

tests/fuzzing/src/fuzz_command_registry.c
```diff
@@ -0,0 +1,57 @@
+#include "fuzz_command_registry.h"
+
+#define FUZZ_COMMAND(group_name,         \
```
cedelavergne-ledger: Not clear how/where are used those defines...

tests/fuzzing/README.md
```diff
@@ -1,141 +1,302 @@
...
+
+- `BOLOS_SDK` pointing at a `ledger-secure-sdk` checkout that ships the
+  `fuzzing/` framework (`fuzzing/cmake/LedgerAppFuzz.cmake`,
+  `fuzzing/scripts/app-campaign.sh`).
+- Clang/LLVM (the Ledger app-builder / dev-tools image ships a supported
+  version), `cmake` ≥ 3.14, `ninja`, `pkg-config`, `libbsd-dev`, `zip`.
+- The first configure fetches the pinned Absolution release; set
+  `LEDGER_FUZZ_ABSOLUTION_LOCAL_DIR=/path/to/absolution` to use a local copy.
+
+## Primary workflow: `app-campaign.sh`
```
cedelavergne-ledger: Not sure I have seen this script??

tests/fuzzing/README.md
```diff
@@ -1,141 +1,302 @@
...
-#### Compile the fuzzers
-
-```bash
-${BOLOS_SDK}/fuzzing/local_run.sh --j=4 --build=1 --BOLOS_SDK=${BOLOS_SDK}
+The `.compat-key` ties the corpus to its build; `app-campaign.sh` and
+`.clusterfuzzlite/build.sh` both refuse to merge a base corpus when the keys
+disagree (a sanitizer change is enough to flip it). Regenerate after any change
+that affects the prefix size or the invariant.
+
+### Findings surfaced by the current corpus
```
cedelavergne-ledger: Not sure we need to keep this in the README??

tests/fuzzing/CMakeLists.txt
```diff
@@ -1,132 +1,133 @@
...
-         ${PLUGIN_SDK_SRC}
-)
+list(APPEND APP_SOURCES ${AB_SDK_SOURCES})
+# main.c: replaced by the fuzz entry point. coverage_sigusr1.c: linked
+# conditionally below. (ui_icons.c is compiled; its generated g_network_icons
+# table is mocked in mock/mocks.c.)
+list(REMOVE_ITEM APP_SOURCES
+    ${APP_SRC}/main.c
+    ${PLUGIN_SDK_SRC}/main.c
+    ${COVERAGE_SIGUSR1_SOURCE})
```
cedelavergne-ledger: Where is the "condition" to compile/link or not the file `overage_sigusr1.c`?

tests/fuzzing/README.md
```diff
@@ -1,141 +1,302 @@
...
+    generate_seed_corpus.py   ← seed generator driven by the command registry
+    fuzz_command_registry.py  ← Python view of the canonical INS table
+
+  macros/
+    add_macros.txt            ← extra -D defines for the fuzz build
+    exclude_macros.txt        ← -D defines removed from the app Makefile defines
 ```
 
-#### Run the fuzzer
+## Production-parity dispatch
```
cedelavergne-ledger: Isn't it something to be stored in the SDK? This seems to be quite generic?

tests/fuzzing/README.md
```diff
@@ -1,141 +1,302 @@
...
-```
 
-### Compilation
-
-```bash
-docker run --rm --privileged -e FUZZING_LANGUAGE=c -v "$(realpath .)/tests/fuzzing/out:/out" -ti fuzz-ethereum
+# Compile inside the container
+mkdir -p tests/fuzzing/out
+docker run --rm --privileged -e FUZZING_LANGUAGE=c \
+    -v "$(realpath .)/tests/fuzzing/out:/out" -ti fuzz-ethereum
```
cedelavergne-ledger: Where/how is generated this docker container?
What about the compile command?

tests/fuzzing/fuzz-manifest.toml
cedelavergne-ledger, line 1: I don't understand this file, and I am not sure it is easy to know what we should do (from scratch, or for updates).....

.clusterfuzzlite/build.sh
cedelavergne-ledger, line 1: This script seems very (too much) complex.
How can we change the way it is constructed to encapsulate the maximum of common parts in the SDK?

src/features/sign_message/cmd_sign_message.c
N3ur0sis: This tripped a UBSan `implicit-integer-truncation` warning (`U4BE()` is `uint32_t` → `uint16_t`) during fuzzing, no real-world impact. Happy to revert to the original with an explicit cast:

`signMsgCtx->msg_length = (uint16_t) U4BE(*data, 0);`

tests/fuzzing/mock/mocks.c
```diff
@@ -0,0 +1,187 @@
...
+ * (recognised by the interceptor) with a compiler barrier preserving the
+ * "do not optimise the clear away" guarantee. Wired via -Wl,--wrap=explicit_bzero;
+ * production keeps the libc / SDK implementation. */
+void __wrap_explicit_bzero(void *s, size_t n) {
+    if (s == NULL) return;
+    memset(s, 0, n);
+    __asm__ volatile("" ::: "memory");
+}
+
+cx_err_t cx_ecdomain_parameters_length(cx_curve_t cv, size_t *length) {
```
cedelavergne-ledger: Why those `cx_XXX` mocks not directly in the SDK framework (potentially with `weak` keyword)? Because I guess they would be needed, in the same way, for lots of Apps?

tests/fuzzing/mock/mocks.c
```diff
@@ -0,0 +1,187 @@
...
+
+#ifdef PRINTF
+#undef PRINTF
+#endif
+int PRINTF(const char *format, ...) {
+    (void) format;
+    return 0;
+}
+
+bool tlv_enforce_u8_value(const tlv_data_t *data, uint8_t expected) {
```
cedelavergne-ledger: Same here: Why do we need a mock here? Why not in the SDK framework?

tests/fuzzing/mock/mocks.h
```diff
@@ -1,6 +1,25 @@
...
+
+/*
+ * Storage for the symbols the SDK harness requires (the framework owns the
+ * logic, the app owns the storage; defined in mocks.c). See the "Mocks" page of
+ * the SDK Fuzzing Framework docs for what each one is and how it is used.
+ */
+extern try_context_t fuzz_exit_jump_ctx;  // setjmp landing pad for app teardown
+
+#define FUZZ_CTRL_SIZE 16
+extern uint8_t fuzz_ctrl[FUZZ_CTRL_SIZE];  // harness control bytes
```
cedelavergne-ledger: Could those "default" init be also hidden in the SDK?

tests/fuzzing/mock/scenario_layout.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+/*
+ * Prefix layout offsets for this app, auto-patched after each build by the
+ * SDK's update-scenario-layout.py. See the Fuzzing Framework docs.
+ */
+#define SCEN_PREFIX_SIZE 231
```
cedelavergne-ledger: Same here: to be hidden in the SDK framework if possible

tests/fuzzing/src/fuzz_dispatcher.c
```diff
@@ -0,0 +1,100 @@
...
+#include "tlv_apdu.h"        // tlv_from_apdu()
+#include "app_mem_utils.h"   // APP_MEM_FREE_AND_NULL()
+#include "mem_utils.h"       // app_mem_init()
+
+#include "fuzz_tlv_config.h"
+#include "fuzz_harness.h"
+
+extern const chain_config_t g_fuzz_chain_config;  // mock/app_globals.c
+
+size_t LLVMFuzzerCustomMutator(uint8_t *data, size_t size, size_t max_size, unsigned int seed) {
```
cedelavergne-ledger: Could it be hidden also?

tests/fuzzing/src/fuzz_dispatcher.c
```diff
@@ -0,0 +1,100 @@
...
+    return fuzz_ethereum_custom_mutator(data, size, max_size, seed);
+}
+
+void fuzz_app_reset(void) {
```
cedelavergne-ledger: The default function from `app_context.c` couldn't be used here?

tests/fuzzing/CMakeLists.txt
```diff
@@ -1,132 +1,129 @@
...
+    ${LEDGER_FUZZ_TLV_MUTATOR_SOURCE})
+
+# ── Per-feature targets ─────────────────────────────────────────────────────
+# Small direct harnesses for the entries a single fuzzed APDU cannot reach on
+# its own: the internal plugins (driven through the plugin interface, not an INS)
+# and the stateful/library parsers (eip712, generic tx parser, calldata). Each
+# drives one module; src/fuzz_utils.c resets the app (heap + globals) per
+# iteration.
+set(ETH_FEATURE_HARNESSES
+    fuzz_calldata
```
cedelavergne-ledger: Do we still need to keep individual harness now with this new framework?

tests/fuzzing/mock/mocks.c
```diff
@@ -0,0 +1,58 @@
...
+ * Nano builds have no such array. */
+#ifdef SCREEN_SIZE_WALLET
+#include "net_icons.gen.h"
+const network_icon_t g_network_icons[10] = {0};
+#endif
+
+/* Link-time mock: the production function triggers an NBGL review flow that
+ * blocks on user interaction; the no-op lets the fuzzer reach the
+ * post-validation path. */
+void ui_display_safe_account(void) {
```
cedelavergne-ledger: Strange to have only a single "UI" related function in the mocks?

tests/fuzzing/CMakeLists.txt
```diff
@@ -1,132 +1,59 @@
...
+
+ledger_fuzz_collect_include_dirs(APP_INCLUDE_DIRS
+    ${APP_SOURCE_DIR}/src
+    ${APP_SOURCE_DIR}/ethereum-plugin-sdk/src)
+list(APPEND APP_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/mock)
+
+# FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION is the shared app/SDK sentinel for
+# fuzz-only relaxations. HAVE_* / APPNAME / ... arrive transitively from
+# secure_sdk (make list-defines + macros/add_macros.txt).
+function(eth_add_fuzz_target NAME INVARIANT)
```
cedelavergne-ledger, line 37: Could/Should we have such a generic function in the framework?

tests/fuzzing/README.md
```diff
@@ -1,141 +1,103 @@
...
+`tests/fuzzing/` (the SDK default is `fuzzing/`). Pass a run name (or omit it for
+a timestamp); outputs land in `.fuzz-artifacts/<name>/`.
+
+## Targets
+
+| Target | What it drives |
+|--------|----------------|
+| `fuzz_app`    | A sequence of fuzzed APDUs through the real `apdu_parser()` → `handleApdu()` path |
+| `fuzz_plugin` | One internal plugin through the production `eth_plugin_call()` sequence |
+| `fuzz_parser` | The generic tx parser, EIP-712 and the calldata store, at their own entry points |
```
cedelavergne-ledger, line 31: Is is pertinent to keep it separately? Is it not already included in the 1st `fuzz_app`?

tests/fuzzing/mock/mocks.c
N3ur0sis: Good question and it turned out the right number is zero, not more. The SDK fuzz framework mocks the whole NBGL layer (fuzzing/mock/nbgl/nbgl_use_case.c), so i don't think app UI need app-side mocks. This one was stale.

tests/fuzzing/CMakeLists.txt
N3ur0sis: There's some app-specific stuff the SDK can't know (include dirs, defines, source glob), so we'll always need a wrapper here. But we could move the ledger_fuzz_bootstrap_file + per-target invariant into the SDK

tests/fuzzing/README.md
N3ur0sis: just ran each target alone (180s each)  and 18 files that are covered by fuzz_parser are at 0% under fuzz_app. for example all file under `src/features/generic_tx_parser/` . the `gtp_*` code is driven through descriptor rather than an INS, so APDU can't get there.  You did catch an inacuracy on the readm though, it justifies the fuzz_parser by the EIP-712 hashers and those are reachable from fuzz_app.

cedelavergne-ledger (approved):

bboilot-ledger (approved):

PR #1060, opened 2026-06-15
fix(eip712): correct message hashing for arrays of structs

LB-thomasp:
## Summary

EIP-712 message hashing (`hashStruct`) is computed incorrectly for several
**array-of-struct** shapes in `src/features/sign_message_eip712/path.c`. In each
case the device signs a digest that does **not** equal the EIP-712 value that
[ethers v6](https://docs.ethers.org/v6/) (`TypedDataEncoder`) and
[@metamask/eth-sig-util](https://github.com/MetaMask/eth-sig-util) (`hashStruct`
V4) **independently agree on**, so a signature produced for such a payload is not
a valid signature over the order the user believes they approved.

**Five** hashing defects are fixed. Each is a distinct structural condition under
which `develop` mis-hashes an array of custom structs (fixture tag in parens,
matching the `eip712_input_files/` fixture names):

- **D1 — rate-boundary element drop** *(bugA)*: an array element struct whose
  encoding length is an exact multiple of the 136-byte keccak-256 rate is judged
  "empty" and dropped from the fold.
- **D2 — nested fixed-array corruption** *(bugB)*: nested fixed arrays of structs
  (depth ≥ 2, e.g. `T[2][2]`) have their array accumulator corrupted, producing a
  wrong, value-dependent hash. Includes Seaport `BulkOrder`
  (`OrderComponents[2][2]`).
- **D3 — empty array before a struct** *(classC)*: an **empty dynamic array**
  immediately followed by a custom-struct field mis-hashes the following struct
  (its typeHash is dropped), for both scalar-element and custom-struct-element
  empty arrays.
- **D4 — struct-first element in a nested array** *(gapA)*: a nested array
  (depth ≥ 2) whose element struct's **first** field is itself a custom struct
  (e.g. `Elem[2][2]` with `Elem{Child c; …}`) — where the D2 nested-array splice
  meets struct-first handling at **inner** array levels.
- **D5 — empty inner struct-array** *(gapB)*: an **empty** custom-struct array
  occupying an **inner** dimension of a multi-dimensional struct array
  (e.g. `Leaf[][]` with `x = [[]]`).

All five (D1–D5) are **pre-existing** defects — `develop` mis-hashes each, so none
is a regression.

The change also pins **one regression guard** *(classR)* — an array element
struct whose **first** field is itself a custom struct. This shape hashes
**correctly on `develop`** and **stays correct** after the fix; it is **not** a
`develop` defect. It is included because an intermediate iteration of the
array-hashing fix regressed it, and the fixture locks in that the final fix does
not. (D4 is the harder, genuinely-broken cousin: the same struct-first shape at
*inner* levels of a nested array.)

The only product change is `src/features/sign_message_eip712/path.c`.

Every messageHash quoted below was produced by compiling the **real** EIP-712
hashing C against the SDK's streaming keccak host-side and replaying genuine
struct-definition/value APDU streams; "correct" is the value on which ethers v6
and eth-sig-util agree.

---

## D1 — array hashed as empty at the keccak-rate boundary *(bugA)*

`finalize_hash_depth()` decided whether a hashing depth had absorbed any data
with a single test:

```c
hashed_bytes = hash_ctx->hash.blen;
...
return hashed_bytes > 0;
```

In the SDK keccak context (`cx_sha3_t`), `blen` is the **partial-block
remainder** — `total_absorbed_bytes mod 136`, where 136 is the keccak-256 rate.
An EIP-712 `hashStruct` input is `typeHash` (32 bytes) plus 32 bytes per field,
i.e. `32 * (1 + nFields)` bytes. When that total is an exact multiple of 136 —
e.g. a **16-field** struct: `32 * (1 + 16) = 544 = 4 * 136` (also 33, 50, …
fields) — `blen` wraps back to `0` though four full blocks were absorbed. The
element's finalized struct hash is judged "empty" and is **not** folded into the
enclosing array, so the array hashes as if it had no elements.

Representative vector — a fixed array of two 16-field structs (`Elem16[2]`):

| | messageHash |
|---|---|
| device (develop) | `0x7e9e2d650137bf3f28375f5a568de3273420e46f2e0b0cdbb840eb9df362252c` |
| correct (ethers v6 = eth-sig-util) | `0xe798863715aebf4d6ccefb2e63d8d5475e38b1cbfe310f9d62a0b6c1bc385ff2` |

**Fix:** treat a depth as non-empty if it processed any full blocks, not only a
partial one. The keccak header counter records already-absorbed full blocks,
recovering the boundary case:

```c
any_absorbed = (hash_ctx->hash.blen > 0) || (hash_ctx->hash.header.counter > 0);
```

---

## D2 — nested fixed arrays (depth ≥ 2) corrupted *(bugB)*

In `path_new_array_depth()`, the custom-struct branch built the array accumulator
by `memcpy`-ing keccak sponge state from one hashing context into another and
re-initializing several contexts in a back-walk loop. For **nested fixed arrays**
(e.g. `T[2][2]`) this clobbers partially-absorbed sponge state, corrupting the
outer-array accumulator and the first inner element, yielding a wrong,
value-dependent hash. It is depth-driven and independent of field count (a
15-field element, which does *not* trigger D1, still mis-hashes at `[2][2]`).

Representative vector — Seaport `BulkOrder` bulk tree (`OrderComponents[2][2]`):

| | messageHash |
|---|---|
| device (develop) | `0x8edb16b93c2a4333eea649c057175495ec3f40e1b5b6dc9a938a48b12a7e8f39` |
| correct (ethers v6 = eth-sig-util) | `0xd76cba560c44183faf0c6033dc2622d4276c04b8ef3f10b2a0ed78dc61b32b0c` |

**Fix:** do not juggle sponge state. Splice a freshly keccak-init'd (empty) array
accumulator into the context list immediately **below** the element context, via
a new static helper `insert_array_hash_depth_before()` built on the SDK's
`list_insert_before()`. This preserves every absorbed byte and mirrors how the
already-correct dynamic-array path behaves.

---

## Regression guard — struct-first element at a single array level *(classR)*

The D2 splice must target the **element struct's** accumulator. When the
element struct's first field is itself a directly-nested custom struct,
`path_update` descends one level further and pushes a grand-child context, so
`get_last_hash_ctx()` points at the grand-child; splicing there would fold
`keccak(hashStruct(child))` and drop the element struct's own child.

**Fix:** resolve the splice target by the element struct's **path depth**
(`pidx + 1`) via the new static helpers `get_struct_ctx_at_depth()` and
`get_element_struct_ctx()`, then splice the array accumulator immediately below
it with `insert_array_hash_depth_before()`. A depth-anchored lookup is correct at
**every** array level and for both element layouts (scalar-first and struct-first),
so it subsumes this single-level regression-guard case and the inner levels of a nested
array uniformly (see D4 below). Single-level `Elem[]`/`Elem[2]` with a
struct-first element therefore stays correct; this case hashes identically on
`develop` and on the fix:

| `Elem[]`, struct-first element | messageHash |
|---|---|
| device (develop) | `0xd926e9c5f07fde8592b51c29b2b33326f0677c55b7a633ed8a429ee964271d1d` |
| device (fixed) | `0xd926e9c5f07fde8592b51c29b2b33326f0677c55b7a633ed8a429ee964271d1d` |
| correct (ethers v6 = eth-sig-util) | `0xd926e9c5f07fde8592b51c29b2b33326f0677c55b7a633ed8a429ee964271d1d` |

---

## D3 — empty dynamic array immediately before a struct field *(classC)*

After an **empty** dynamic array, the unwind (`path_advance(false)` →
`path_update`) descended into the **following** custom-struct field without
feeding that struct's typeHash, so its `hashStruct` was computed over its encoded
fields alone (missing the typeHash), corrupting the message hash. This affects
both an empty **scalar** array before a struct (`Root{uint256[] a = [], S b}`)
and an empty **custom-struct** array before a struct
(`Batch{Call[] calls = [], Meta meta}`, and the variant where the following
struct's leading field is itself a nested struct).

**Fix:** for an empty scalar array, unwind with `do_typehash = true` so the
following struct receives its typeHash. For an empty custom-struct array, keep
the unwind `do_typehash = false` (the element struct was already descended during
setup) and, after the unwind, repair every struct depth the unwind reopened by
feeding the missing typeHash — covering the immediately-following struct, its
leading nested structs, and a sibling struct reached after the array's enclosing
struct is closed.

Representative vector — `Batch{Call[] calls = [], Meta meta}` (empty
custom-struct array then struct):

| | messageHash |
|---|---|
| device (develop) | `0x25d4f8ff17db19ee970dad75fdbfe3f4241fa43a82d87c6d23420d80bd0d6254` |
| correct (ethers v6 = eth-sig-util) | `0xd9c17bb29bca32d163e1b7fcc452e3f7b86f1e33a19c54664719cd9143e97b67` |

---

## Real-app impact

Verified against the host SUT (device-computed vs. ethers v6 = eth-sig-util):

- **Fixed (was broken):** Seaport **bulk** `BulkOrder` (`OrderComponents[2][2]`)
  — broken on `develop` (D2), now matches.
- **Unaffected (match on both `develop` and the fix):** Safe `SafeTx`/
  `SafeMessage`; Permit2 `PermitSingle`/`PermitBatch`; Seaport **single-level**
  orders and `OrderComponents[2]`; ERC-2612 permit; ENS; CoW. These are
  canonical scalar-first shapes and were never mis-hashed.
- **Newly addressed beyond bulk orders:** generic multicall/batch wallets that
  place an **empty array immediately before a struct field** (D3) now hash
  correctly; and **nested** arrays of custom structs whose element's first field
  is a struct (D4) or whose inner dimension is an empty struct-array (D5)
  — both broken on `develop` — now match the oracle.

---

## D4 & D5 — previously carved-out nested gaps *(gapA / gapB)*

Two **pre-existing** nested-array shapes — formerly carved out under KNOWN
LIMITATIONS because they mis-hashed on `develop` **and** the original 4-class fix
— are now **fixed** by the depth-anchored element-context handling
(`get_struct_ctx_at_depth` / `get_element_struct_ctx` /
`insert_array_hash_depth_before` in `path.c`). They are pre-existing defects, not
regressions; the oracle (ethers v6 = eth-sig-util V4) is unambiguous for both:

- **D4** — a nested array (depth ≥ 2, fixed or dynamic) whose element struct's
  **first** field is itself a custom struct (e.g. `Elem[2][2]` with
  `Elem{Child c; …}`). This is the intersection of the D2 nested-array splice
  and the regression-guard struct-first element handling at **inner** array levels: the
  array accumulator was spliced at the wrong hash context. Now fixed for fixed
  `Elem[2][2]`/`Elem[2][2][2]`, dynamic `Elem[][]`/`Elem[][][]`, and mixed
  `Elem[][2]`.
- **D5** — an **empty** custom-struct array occupying an **inner** dimension
  of a multi-dimensional struct array (e.g. `Leaf[][]` with `x = [[]]`). The
  empty-inner unwind folded `keccak256("")` at an array-accumulator context
  rather than a struct depth. Now fixed for `Leaf[][]`, `Leaf[][2]`, `Leaf[][][]`
  with empty inner dimensions, and mixed `x = [[],[leaf],[]]`.

Both are now **real pass-on-fixed device fixtures** (NN 50..58 under
`eip712_input_files/`, auto-discovered and signed by `test_eip712_new`). Each was
also verified against the **real `path.c`** compiled host-side (the SDK
streaming-keccak SUT): every GAP fixture
**diverges on pristine `develop`** and **matches the oracle on the fixed
source**. The earlier `known_uncovered_*` carve-out fixtures and
`test_eip712_known_uncovered.py` are removed.

---

## Tests

All test additions are **functional** (`tests/functional/`); no product code
other than `path.c` changes.

- **Structural device vectors.** Fixtures auto-discovered by
  `test_eip712.py::test_eip712_new`, which signs each on-device and asserts the
  recovered signer equals the device key. They cover the D1 boundary
  (16/33/50-field, single-level fixed & dynamic, directly-nested, and the minimal
  single directly-nested element), D2 (`Small[2][2]`, `Small[2][2][2]`,
  Seaport `OrderComponents[2][2]`), regression-guard (struct-first element, dynamic &
  fixed), D3 (empty scalar array then struct; empty custom array then
  struct; empty custom array then nested struct; non-empty control) and the
  now-fixed **D4** (struct-first element in a nested `Elem[2][2]`/`Elem[][]`/
  `Elem[][2]`/`Elem[][][]`/`Elem[2][2][2]`) and **D5** (empty inner
  custom-struct array `Leaf[][]`/`Leaf[][2]`/`Leaf[][][]`, plus mixed
  empty/non-empty inner). Real-app controls (Mail `Person[]`, Seaport single
  order, Safe `SafeTx`, Permit2 `PermitSingle`) are included. Every fixture is
  signed on-device and checked by recovering the signer against the `eth_account`
  (Ethereum Foundation) EIP-712 oracle the upstream functional suite already uses.

- **Host differential against the real `path.c`** (development aid, not part of
  this PR). Each fixture's APDU stream was replayed through the real EIP-712
  hashing C compiled host-side (built into `fuzz_eip712` against the SDK streaming
  keccak), comparing the device-C `messageHash` to an independent oracle
  (eth-sig-util V4 = ethers v6). Every bug-trigger vector — D1, D2,
  D3, **D4**, **D5** — diverges on pristine `develop` and matches on
  the fixed source; controls match on both. A 300-seed differential refuzz with a
  nested-custom-array generator (struct-first nested elements and empty-inner
  struct-arrays — the blind spot that hid D4/D5) reported 0 divergences on
  the fixed source.

**Fail-before / pass-after:** every bug-trigger fixture recovers a wrong signer
on `develop` and the correct signer with the fix; every control recovers correctly
on both.

> The committed test surface is `path.c` plus the functional fixtures under
> `tests/functional/eip712_input_files/` (signed by `test_eip712_new`). The host
> differential (`tests/fuzzing/` replay target + `.bin` corpus) is a development
> aid, not part of this PR — see *Differential-testing recommendation* for
> committing a durable fuzz target.

### Verification results

Run on the fixed branch (`fix/eip712-nested-array-hashing`):

- **Functional Speculos** — `test_eip712.py::test_eip712_new` auto-discovers and
  signs all 59 `eip712_input_files/*-data.json` across `verbose_raw ∈ {true,false}`
  and asserts `recover(signature) == device key`: **126 passed / 0 failed** (run on
  both **Nano S+** and **Flex**). The remaining `test_eip712.py` cases (filtering,
  trusted-name, token metadata, calldata/proxy, gondi, batch, v0 legacy):
  **24 passed, 3 conditional-skip, 0 failed**.
- **Host differential vs. real `path.c`** — every fixture's APDU stream replayed
  through the EIP-712 hashing C compiled host-side, compared to the
  eth-sig-util V4 = ethers v6 oracle: **43 / 43 match** on the fixed source; every
  bug-trigger vector (D1/D2, D3, D4/D5) **diverges on `develop`** and
  matches after the fix; controls match on both.
- **On-device (physical Nano S+)** — the fixed app was built and side-loaded onto
  a physical **Ledger Nano S+** and exercised against the Bug-Gallery shapes
  (D1–D5 and the regression guard) and against real NFT bulk-order signatures for
  **Seaport** (`BulkOrder`) and **Payment Processor** (`BulkSaleApproval`); every
  signature recovered to the device key (correct on the fix; the same payloads
  recover to the wrong signer on the released 1.22.1 app).

---

## Device coverage

The only product change is `src/features/sign_message_eip712/path.c`, which is
**device-independent hashing logic** (keccak sponge, the 136-byte rate boundary,
and the struct fold). It is compiled identically for every supported target, so
the defect and its fix apply uniformly to **all** devices the app installs on —
`nanos+`, `nanox`, `stax`, `flex`, `apex_p` (`ledger_app.toml`); the original
Nano S is not a target. This is confirmed three ways: the host differential
compiles the *exact* `path.c` for the host (the hashing math, no device involved);
the functional suite runs on Speculos for **Nano S+** and **Flex** —
architecturally distinct UI stacks (BAGL/buttons vs NBGL/touch) — with identical
results (**126/126** each); and the app was built and **side-loaded onto a
physical Ledger Nano S+** and signed the Bug-Gallery shapes and real Seaport /
Payment Processor NFT bulk orders correctly on hardware. The app builds cleanly
from the same source for the remaining targets (`nanox`, `stax`, `apex_p`), so the
hashing behaviour is identical across them; the bug was never Nano S+-specific.

---

## Interactive reproduction

An interactive Next.js site reproduces every case against a live MetaMask and/or
a Ledger over WebHID, signing and checking signer recovery per shape (Payment
Processor `BulkSaleApproval` merkle trees, Seaport `BulkOrder`, a direct-WebHID
APDU section, and a per-defect gallery): <https://gist.github.com/LB-thomasp/d766bf21259b0535f6b37206c5db09cc>

---

## A note on the added fixtures

The fixtures are validated by the functional streaming path
(`test_eip712_new`), which encodes them with the official test client. Two
fixture-authoring details were required for that path (neither affects any hash
or the bug coverage): each fixture carries an explicit `EIP712Domain` type entry,
and custom struct types are not named with a trailing digit — the test client's
`get_typesize()` parses a trailing digit run as a numeric type size (valid only
for native `uint`/`int`/`bytes`), which would truncate a name like `Elem16` to
`Elem`. The device itself stores and resolves digit-suffixed struct names
correctly; this is purely a test-client encoding constraint.

---

## Differential-testing recommendation

The durable safeguard for this class of defect is a **differential EIP-712
hashing oracle** added to `tests/fuzzing/`: a target that, for each generated
typed-data case, compares the app's streamed `hashStruct` against an independent
reference (eth-sig-util V4 / ethers v6) and flags any divergence. D1, D2,
D3, D4 and D5 are all reachable by such a generator once it emits
arrays of custom structs — including **nested** `C[][]` / `C[2][2]` with
struct-first elements and **empty inner** struct-arrays, which a
single-array-level grammar misses (this blind spot is exactly what hid D4/
D5). Pinning this as a CI fuzz target would catch the whole family
automatically.


LB-thomasp:
removed the parity tests to resolve CI assertion failures

Failures stemmed from a test file that used hashes generated with typescript signing utilities such as ethers and MetaMask eth-sig-util to ensure hash parity across signing platforms. The issue with trailing digits truncating a type name failed one set of tests locally and only that set was re-run after changing the type names.

Since those tests are redundant to the `eth_account` based tests, would incur maintenance overhead or additional CI to generate, and were primarily used for local validation across multiple EIP712 hashing packages due to the origin of this PR being reports from a user that their MetaMask was erroring when signing a message on their Ledger... it's best to drop the new assertion tests and leave the expanded input files for testing for coverage.

